PR #30, opened 2023-03-28
feat: 0.4 - 月夜の音楽会 (Moonlit Night Concert)

SaltyAom:
Version name after [the opening of "The Liar Princess And The Blind Prince"](https://youtu.be/UdBespMvxaA).
Branch name after [Dorothy's Magic Theorem](https://youtu.be/fk38p6N92qg).

---

This will be not a significant update to introduce a new feature but an internal improvement, and might contain some breaking change for the future:

Featuring:
- AoT compile
- TypeBox 0.26
- Validate response per status instead of union
- `if` for conditional route


SaltyAom:
LGTM 👍

1d7b865f006dfcb83a8c157901ac68c0188c459e_2_690x460

PR #33, opened 2023-04-09
Add support for "parse" lifecycle event for websockets.

ntzrbtr:
One of the clients in my project tries to open a websocket connection with the content type set to "application/json", but without sending a request body. This results in an error in the compiled handler (line 71 in compose.ts), so in order to fix this, I would need support for the "parse" lifecycle event also for websockets.

The other solution would be to do a more fine-grained handling of errors in the generated code in compose.ts, but for now I like the simplicity of the approach with only one try-catch block for everything.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1846,6 +1846,7 @@ export default class Elysia<Instance extends ElysiaInstance = ElysiaInstance> {
 				return 'Expected a websocket connection'
 			},
 			{
+				parse: options.parse,
 				beforeHandle: options.beforeHandle,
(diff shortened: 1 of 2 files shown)


SaltyAom:
Sorry for the slow response, I had a vacation the last week and didn't have a chance to check the mail.

Back to the topic, right, sounds like the problem with generated code, we will try to improve it in the later version to be more fine-grained in the future, but this is good enough for now.

SaltyAom:
Should be available in 0.4.9

PR #47, opened 2023-05-26
return an instance of the class in the handler

shad0w-jo4n:
Before that PR you could only return anonymous objects:

```ts
app.get('/', () => {
  return {
    message: 'Hello, world!',
  };
});
```

After that PR you can return an instance of the class:

```ts
class MessageResponse {
  constructor(public message: string) {}
}

app.get('/', () => {
  return new MessageResponse('Hello, world!');
});
```

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -42,21 +42,27 @@ export const mapEarlyResponse = (
 				set.headers['Set-Cookie']
 			)
 
-		switch (response?.constructor?.name) {
-			case 'String':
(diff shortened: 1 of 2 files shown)


SaltyAom:
Hi, thanks for the contribution.

However, this PR introduces some changes to `switch-case` mapping to `if` which has a performance impact on the case.

I create a replication to add this feature with d6e6830, released with Elysia 0.5.8

Thanks for your contribution.

PR #85, opened 2023-08-09
fix: groups with prefixed instance

Attk4:
Hi!

As i already mentioned this on Discord, groups do not work correctly with the new plugin model introduced in 0.6, this fixes that issue.

image

There is still an issue with the inferred type when we use a plugin prefix, but hopefully we can fix that too in the near future.


SaltyAom (approved):

SaltyAom:
Hi, thanks a lot for the PR! I didn't realize this was occurring lately.

PR #92, opened 2023-08-22
fix(websocket/types): pass full instance and extend context with its …

### LINKED ISSUE #59: WebSocket beforeHandle unable to access use'd plugins / state / derive's
I'd like to have the ability to check for certain cookies before allowing a WebSocket connection, as well as have access to the Elysia context. Here's some demo code:

```ts
import { Elysia, ws } from 'elysia'
import cookie from '@elysiajs/cookie';

new Elysia()
	.use(cookie())
	.use(ws())
	.state('count', 0)
	.ws('/ws', {
		beforeHandle({ cookie, store }) {
			if (!cookie['authToken']) return new Response(null, { status: 401 });

			console.log(store.count);
		}
	})
	.listen(8080)
```

When I run `tsc`, I get these errors for `cookie` in the params and `store.count`:
```
index.ts:9:18 - error TS2339: Property 'cookie' does not exist on type '{ request: Request; headers: Record<string, string | null>; query: Record<string, unknown>; params: Record<never, string>; body: unknown; store: {}; set: { headers: Record<...>; status?: number | undefined; redirect?: string | undefined; }; } & Record<...>'.

9   beforeHandle({ cookie, store }) {
                   ~~~~~~

index.ts:12:22 - error TS2339: Property 'count' does not exist on type '{}'.

12    console.log(store.count);
                        ~~~~~


Found 2 errors in the same file, starting at: index.ts:9
```

Ideally, I should be able to access anything available on the context in these WebSocket handlers as I would in regular HTTP endpoints.

I'm currently using Bun `0.6.8`, and `bun pm ls` outputs
```
├── @elysiajs/cookie@0.5.2
├── bun-types@0.6.9
├── elysia@0.5.19
└── typescript@5.1.3
```

EDIT: This might not be relevant for right now, but I'd like to see the notion of a derive in the context of WebSockets, e.g. some way to add data to the "ws" objects passed into open / message / close. This'd be real handy if I wanted to attach certain data, such as user IDs.

hesxenon:
Since the `ElysiaWsContext` hasn't been extended with the `Instance['request']` websocket handlers didn't have access to derivations and such.

This PR simply passes the full instance instead of just the schema definitions and extends the context with its request type

fixes #59

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2797,7 +2797,7 @@ export default class Elysia<
 			? ElysiaWSOptions<
 					`${BasePath}${Path}`,
 					Schema,
-					Instance['meta']['defs']
+					Instance
 			  >
 			: never
 	): Elysia<
--- a/src/ws/types.ts
+++ b/src/ws/types.ts
@@ -70,23 +70,23 @@ export type TransformMessageHandler<Message extends WSTypedSchema['body']> = (
 
 export type ElysiaWSContext<
 	Schema extends WSTypedSchema<any> = WSTypedSchema<never>,
-	Definitions extends ElysiaInstance['meta']['defs'] = {},
+	Instance extends ElysiaInstance = ElysiaInstance,
 	Path extends string = never
 > = ServerWebSocket<
 	Context<{
-		body: UnwrapSchema<Schema['body'], Definitions>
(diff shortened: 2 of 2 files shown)


SaltyAom:
This is super helpful!
Thanks a lot!

PR #103, opened 2023-09-06
Release 0.7: Miyu (codename)

SaltyAom:
Release name: ? (decided, not revealed yet)

Codename [**Kasumizawa "Miyu"**](https://bluearchive.fandom.com/wiki/Kasumizawa_Miyu) from Blue Archive
<a href=https://bluearchive.fandom.com/wiki/Kasumizawa_Miyu>
Kasumizawa Miyu, small anime girl in blue student sailor uniform in blue holding a sniper rifle gun in shy manner
</a>

- [x] Rewrite Type
    - Simplified type
    - Increase type inference performance
    - Measured improvement speed by ~13x
    - [x] Lifecycle
    - [x] State & Decorate
    - [X] Model & Error
    - [x] Event
    - [x] Runtime
    - [x] End-to-end type safety
    - [x] Websocket
    - [x] test case
- [X] Rewrite WebSocket
    - Strict validation for both runtime and type
    - Simplify Web Socket process
    - [x] Remove path mapping and use the compose function via hoisting instead
    - [x] Integrate with state and decorate fully type-safe
    - [x] Runtime
    - [x] test case (contributed by @bogeychan)
- [X] Trace
    - Monitoring via life cycle and unit
    - Measure performance between life-cycle and sub-unit (array)
    - [x] `.trace`
    - [x] compose
    - [x] `request`, `request.unit`
    - [x] `parse`, `parse.unit`
    - [x] `transform`, `transform.unit`
    - [x] `beforeHandle`, `beforeHandle.unit`
    - [x] `handle`
    - [x] `afterHandle`, `afterHandle.unit`
    - [x] `response`, `response.unit`
    - [x] `error`, `error.unit`
    - [x] deciding to integrate throw catch error or not
    - [x] test case
- [x] Reactive Cookie
    - Migrate cookie plugin to core without performance drop
    - Simplified Getter Setter #98
    - [x] get, set
    - [x] delete
    - [x] append
    - [x] signed, unsigned
    - [x] expire
    - [x] secret, rotation
    - [x] compose
    - [x] validation
    - [x] config
    - [x] test case 
- [X] Mapper Method & Affix #83
  - [x] affix
  - [x] prefix, suffix
  - [x] test case 
- [X] New Type
    - [x] Nullable
    - [x] MaybeEmpty
- [x] Better Test Case and Coverage
    - Add more tests for better coverage
    - Improve contribution testing
    - [x] Core
      - [x] Module
      - [x] Handle Error
    - [x] Handler
[...]

PR #116, opened 2023-09-10
Made reusable route types!

ArcherHume:
Hi!

I noticed in `index.ts` there is a repeated use & recreation of one type for each route method, this has now been moved into `types.ts`, reducing the length of the file by ~1000 lines & significantly increasing readability.


SaltyAom:
Thanks a lot for your contribution, but sadly I'll not merge this for 2 reasons.

1. We had a reusable route type before, it's easy to maintain but causes us type degradation, by a lot. Yes, we are looking for a way to have a reusable type route.
2. We already already done that in the next release, see #103

If you have more question, feels free to ask.

bharat5604 (approved):

SaltyAom:
As there's no on going discussion and this is essentially solved in #103, I'm closing the PR.

PR #132, opened 2023-09-12
Implement JsonString

### LINKED ISSUE #131: [Feature Request]: Accept "JsonString" custom type
It's common, while working with multipart/form-data requests to send fields with encoded json, and currently, parsing these fields is a pain since you'll need to treat them as strings and on the handler itself you'll need to do the validation.

See: https://discord.com/channels/1044804142461362206/1044807567022493766/1151065279166689310

Current Example:

```ts
const app = new Elysia()
    .post("/create-user", () => {}, {
         picture: t.File(),
         name: t.String(),
         // It's an object, but unfortunately we need to treat them as strings and parse it inside the handler
         address: t.String() 
     })
```

Suggested API:
```ts
const app = new Elysia()
    .post("/create-user", () => {}, {
         picture: t.File(),
         name: t.String(),
         address: t.JsonString(
            t.Object({
                country: t.String(),
                address1: t.String()
            })
         ) 
     })
```

brunoeduardodev:
Closes #131

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -121,6 +121,10 @@ export const findElysiaMeta = (
 	parent = ''
 ) => {
 	if (schema.type === 'object') {
+		if (schema.elysiaMeta === 'JsonString' && type === 'JsonString') {
+			found.push(parent)
+		}
+
 		const properties = schema.properties as Record<string, TAnySchema>
 		for (const key in properties) {
 			const property = properties[key]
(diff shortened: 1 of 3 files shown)


SaltyAom:
Thanks for your PR. I use this as a base to implement JsonString in #103 using Encode and Decode instead.

This helps me set a starting point to implement the feature by a lot, so I'm closing with a credit.

PR #144, opened 2023-09-14
updated error constructors to take in a custom error message

brayden-wong:
I noticed that the default error was just left to the error message was set as a default to whatever message you send, however it would be nice if out of the box, we as developers could add a custom message to those error types(all excluding the validation error). So I implemented an overloaded constructor to take in this option and implemented it with Elysia's base error types and added tests to verify they work.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


SaltyAom (approved):

SaltyAom:
Hi, thanks for your this feature, can you help me resolve the lock file, thanks.

brayden-wong:
I'm not sure how I can resolve this. All I did was do a `bun install`, and just assumed that we will take the most recent bun.lockb file

PR #149, opened 2023-09-14
feat: support additional status codes in redirects

### LINKED ISSUE #142: Feature Request: Support additional 3XX status codes in redirects
Currently cannot redirect to another route with an HTTP response code other than 302. 

For example, here I am trying to set a 303 so when I issue a DELETE request to `/events/:id` it will then issue a GET request on redirection to `/events`

 ```ts
const app = new Elysia()
  .use(html)
  .get("/events", getEvents)
  .delete("/events/:id", async ({ params, set }) => {
    await deleteEvent(params.id);
    set.status = 303
    set.redirect = '/events'
  })
```

But it seems like a 302 response code is hardcoded with a `set.redirect`:
https://github.com/elysiajs/elysia/blob/741ffe7659cf049c8627d30b91cd1f05cfbab53e/src/handler.ts#L32C3-L35

CKonv:
Hi guys, I wrote a small fix to allow a custom status to be set on redirects. This would close #142 

Please let me know if I have missed anything, as this is my first OSS contribution 😄


arsaizdihar:
I think it would be better if instead of checking the 200 status code, it checks whether the status code already 3xx or not. Only change to 302 if it is not.

CKonv:
Hi, thank you for your input. I also thought about doing it that way, but initially decided against it, because as a user, I would expect any explicitly set status codes to remain, even if they are in the wrong range.

SaltyAom (approved):

PR #152, opened 2023-09-15
fix: bad code in Internal server error class

l2ysho:
This fixes issue when NotFoundError and InternalServerError has the same code 'NOT_FOUND'.


brayden-wong:
I fixed this in my PR https://github.com/elysiajs/elysia/pull/144

Issue #195, opened 2023-09-20
[bugreport] TypeError: undefined is not an object (evaluating 'this.raw')


eelkevdbos:
When destructuring the ws param in `open(ws) { }`, the method starts throwing: `TypeError: undefined is not an object (evaluating 'this.raw')`. It looks vaguely related to #174.

My environment

```json
{
  "devDependencies": {
    "bun-types": "^1.0.1"
  },
  "peerDependencies": {
    "typescript": "^5.2.2"
  },
  "dependencies": {
    "@elysiajs/html": "^0.7.0",
    "@kitajs/html": "^2.2.2",
    "@kitajs/ts-html-plugin": "^1.0.2",
    "elysia": "^0.7.3"
  }
}
```
Snippet to reproduce:

```ts
new Elysia()
  .ws('/ws/:channel', {
    params: t.Object({
      channel: t.String(),
    }),
    open({ subscribe, data: { params } }) {
      subscribe(params.channel) // throws TypeError
    },
  })
```

Snippet that does not throw:

```ts
new Elysia()
  .ws('/ws/:channel', {
    params: t.Object({
      channel: t.String(),
    }),
    open(ws) {
      ws.subscribe(ws.data.params.channel)
    },
  })
```

erra0002:
This is expected behavior, destructuring assignment is not "supported" in javascript for class methods like `subscribe`, because the context and therefore "this" is lost. You have to use `ws.subsribe(...)`

PR #201, opened 2023-09-21
fix: allow WebSocket destructuring

bogeychan:
Closes https://github.com/elysiajs/elysia/issues/195

But idk if it's worth it, as someone mentioned here:: https://github.com/elysiajs/elysia/issues/195#issuecomment-1728545341

### DIFF
--- a/src/ws/index.ts
+++ b/src/ws/index.ts
@@ -35,63 +35,74 @@ export class ElysiaWS<
 		this.validator = raw.data.validator
 	}
 
-	publish(
-		topic: string,
-		data: Route['response'] = undefined,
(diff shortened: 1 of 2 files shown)


eelkevdbos:
I kind of like it, but I would also be ok with letting it rest. It's just that the docs encourage destructuring in a lot of places and it wasn't clear to me from the type signature that it would fail during runtime because it was a class method as described in the comment.

SaltyAom:
I think this is fine, and is a nice addition.

Creating a new function usually doesn't affect performance much, so I'd rather have this.

Thanks you for your contribution.

Issue #153, opened 2023-09-15
feat: Cover base error cases with classes.


l2ysho:
Elysia now support InternalServerError, NotFoundError, ParseError and ValidationError which covers 400, 404 and 500 error types. Lets add also 401, 403 and 409 at least to cover all basic situations.

PR #223, opened 2023-09-24
feat: add support for 401, 403 and 409 errors

l2ysho:
This PR closes https://github.com/elysiajs/elysia/issues/153

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -14,11 +14,14 @@ export const ERROR_CODE = Symbol('ErrorCode')
 export const isProduction = (env?.NODE_ENV ?? env?.ENV) === 'production'
 
 export type ElysiaErrors =
+	| ConflictError
+	| ForbiddenError
 	| InternalServerError
+	| InvalidCookieSignature
 	| NotFoundError
 	| ParseError
+	| UnauthorizedError
 	| ValidationError
-	| InvalidCookieSignature
 
 export class InternalServerError extends Error {
 	code = 'INTERNAL_SERVER_ERROR'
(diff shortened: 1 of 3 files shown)


yfrans:
After searching for this feature with no luck, I developed a small helper library called elysia-http-error which allow throwing HttpError messages more easily.
You can check it out in NPM registry: https://www.npmjs.com/package/elysia-http-error

This helped me a lot, for example when throwing this error from a derived method which checks jwt then returning the payload to be used later in a route, it allows using the type returned from the derived method without the need for null checking.

l2ysho:
@yfrans looks good, but I think it is not big deal to implement it directly to elysia. Maybe some response from other contributors or code owners would be useful cc @SaltyAom

yfrans:
I prefer throwing exception on errors instead of using status codes.
It allows better code flow because error is error and should break the response flow (again, my opinion).
Generic HttpError class is fine, but using a class with static initializers for common errors is more easy to handle.
But it is your library, we can only advice 😀

l2ysho:
For me the same as for @yfrans, it is more clear for me to throw specific error right away
```javascript
if (!user) throw UnauthorizedError('Error message')
if (!permissions) throw ForbiddenError('Error message')
if (!alreadyExists) throw ConflictError('Error message')
if (!record) throw NotFoundError('Error message')

``` 

but it make sense to have one generic elysia error and handle it in onError handler

```javascript
if (!customCondition) {
      throw new ElysiaApiError(409, 'Error message')
}
```

So if you choose you prefer way I can update PR.

yfrans:
@l2ysho I think using http code like "409", "401", etc. is similar to setting the res.status (without considering the advantage of throwing an exception).
That's why I added the initializers for each http status code.

gabbrieu:
Any news here?

zmzlois:
> For me the same as for @yfrans, it is more clear for me to throw specific error right away
> 
> ```js
> if (!user) throw UnauthorizedError('Error message')
> if (!permissions) throw ForbiddenError('Error message')
> if (!alreadyExists) throw ConflictError('Error message')
> if (!record) throw NotFoundError('Error message')
> ```
> 
> but it make sense to have one generic elysia error and handle it in onError handler
> 
> ```js
> if (!customCondition) {
>       throw new ElysiaApiError(409, 'Error message')
> }
> ```
> 
> So if you choose you prefer way I can update PR.

The time for a js/ts framework to have match case. Will Elysia be the first?

SaltyAom:
#284 introduces `error` function to return a custom error without using `set`.

This function provides type inference for use later with Eden.

With the same intent functionality, I'm closing this PR.

Issue #232, opened 2023-09-27
Schema with Type.Record crash app


timnghg:
Failed test case
https://github.com/elysiajs/elysia/pull/233/commits/bdcbef9573f6d79a8bfce5825717999cca081fae#diff-038c1441ffe466797ddc781688224e001f6f39839cb6e3405d8d2c2f8ab35901

![image](https://github.com/elysiajs/elysia/assets/140995877/4daad32a-2f14-4fb8-879e-e0d241d2acec)

SaltyAom:
Fixed with 0.7.17.

Feels free to reopen this issue if the problem still persists.

PR #233, opened 2023-09-27
fix: #232 prevent app crashes because of schema type record

timnghg:


SaltyAom:
Fixed with another PR, but thanks for your contribution.

PR #227, opened 2023-09-26
add "http://" to Elysia's running message

ramdhanyA:

PR #237, opened 2023-09-28
feat: log a valid link to starting a server with lazy-module

fredericoo:
Super quick fix: by including the protocol (which is hardcoded to `http` as we don’t have access to it, and will likely not ever be a problem as HTTP rewrites would have this covered too)

CleanShot 2023-09-28 at 10 08 30@2x

This change would make it so that the output link on the console would be clickable via <kbd>Cmd</kbd>+Click ◡̈


fredericoo:
Duplicate of https://github.com/elysiajs/elysia/pull/227

PR #265, opened 2023-10-12
derive placement changed to allow handling after other transforms

abhi-slash-git:
Modifying how and where derive transform takes place. This allows derive to trigger after the validations have taken place. This allows for cases where you want validators to reject before deriving a value into decorators or stores.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -864,26 +864,26 @@ export const composeHandler = ({
 	endParse()
 
 	if (hooks?.transform) {
+		const nonDeriveTransforms = hooks.transform.filter(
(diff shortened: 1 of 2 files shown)


SaltyAom:
The problem with this is that `derive` is expected to behave like `transform`, which means it would be run before the validation process.

This means that we may have to introduce a new API rather than changing how `derive` works which may cause unexpected breaking change that is hard to trace.

Please leave open until I announce the said API.

abhi-slash-git:
I'll make an issue, I like the idea of adding another API. Thanks for reviewing.

PR #295, opened 2023-11-06
Allow 'true' as option for Record-type params

adicco:
This would enable the following syntax:

```ts
// Let's assume this returns an array of paths like
// ['/some/:path', '/some/other/route']
// And that the files in the pages/ folder all export some Handlers
const dynamicRoutes = glob("pages/**/index.tsx");

const app = new Elysia();

dynamicRoutes.map(path =>
  app.
    .get(path, (ctx) => {
        // ctx.params is of type Record<string, string>
      },
      {
        params: true
      }
    )
)
```

While at the moment you get:.

```ts
const app = new Elysia()
  .get('/whatever/route/with/maybe/params' as string,
  (ctx) => {
    // ctx.params is of type never
  })
```

Which would greatly enable the flexibility of Elysia, particularly when it comes to building frameworks on top of it. Handlers can then have their own inner validation of the params.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -168,7 +168,9 @@ export type UnwrapRoute<
 	> extends infer A extends Record<string, any>
 		? A
 		: undefined
-	params: UnwrapSchema<
+	params: Schema['params'] extends true
+	?  Record<string, string>
+	: UnwrapSchema<
 		Schema['params'],
 		Definitions
 	> extends infer A extends Record<string, any>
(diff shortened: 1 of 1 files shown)


adicco:
Actually, I am thinking that exporting handlers might not be the best way to proceed, and I should rather export full Elysia instances, which I can then dynamically import and `use`.

SaltyAom:
I'm not sure what you are trying to do, how is this difference from [Wildcard](https://elysiajs.com/concept/route.html#wildcard)?

adicco:
> I'm not sure what you are trying to do, how is this difference from [Wildcard](https://elysiajs.com/concept/route.html#wildcard)?

Wildcard is typed as a TS literal and so can match your "*" type inference. If I use a `string` variable it will not infer the proper params and type it to `never`.

But I think I have found another way to do what I wanted, which uses dependency injection and is more in line with Elysia's philosophy. I was coming more from a Next.js approach but this does not seem to fit too well.

PR #314, opened 2023-11-16
Fix Unable to dereference schema with  'undefined' when using t.Ref

dungps:
When I define a typebox object with `$id` at `Elysia.model` and I use `t.Ref` in hooks then it will throw the error `Unable to dereference schema with $id 'undefined'` when starting the server.

This PR will add model references to TypeCompiler.Compile to fix this issue.

Error example:

```
new Elysia()
    .model({
        user: t.Object(
            {
                id: t.Number()
            },
            {
                $id: "user"
            }
        )
    })
    .get(
        "/",
        () => ({
            success: true,
            data: {
                id: 1
            }
        }),
        {
            response: {
                200: t.Object({
                    success: t.Boolean(),
                    data: t.Ref("user")
                })
            }
        }
    )
    .listen(3000)
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -168,7 +168,7 @@ export const getSchemaValidator = (
 			Code: () => ''
 		} as unknown as TypeCheck<TSchema>
 
-	return TypeCompiler.Compile(schema)
+	return TypeCompiler.Compile(schema, Object.values(models))
 }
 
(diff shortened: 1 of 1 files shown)


SaltyAom:
Error occurs from sensitive header case, will fix that on local.

PR #330, opened 2023-11-23
fix: preserve query params for mounted handler

### LINKED ISSUE #329: mount() does not preserve query params
When mounting a handler via `mount()`, the received URL is missing whole `query` from original URL:
[https://github.com/elysiajs/elysia/blob/757357f28749674844fc848e4ddd715134f7714f/src/index.ts#L1715](https://github.com/elysiajs/elysia/blob/757357f28749674844fc848e4ddd715134f7714f/src/index.ts#L1715)

silvio2402:
Query parameters weren't preserved for a mounted handler. This was because the new request url passed to the handler was assembled using only the hostname and path.

This fixes  #329

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -26,7 +26,7 @@ import {
 	mergeLifeCycle,
 	filterGlobalHook,
 	asGlobal,
-	getHostname
+	replaceUrlPath
 } from './utils'
 
 import {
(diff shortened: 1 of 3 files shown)


Arthurmtro (approved):
Good job, it fixes it well

dodas (approved):

PR #380, opened 2023-12-26
fix: display the correct error message when calling .listen outside of bun

paperdave:
There was code to display this message but it didn't get hit because a second check which wasn't written correctly errors out first.

```diff
- ReferenceError: Bun is not defined
+ Error: .listen() is designed to run on Bun only. If you are running Elysia in other environment please use a dedicated plugin or export the handler via Elysia.fetch
```

Tested manually in Node. Attempted to test in Deno but their `--unstable-byonm` flag didn't seem to work, though it is possible to load the library through npm.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3973,7 +3973,10 @@ export default class Elysia<
 		options: string | number | Partial<Serve>,
 		callback?: ListenCallback
 	) => {
-		if (!Bun) throw new Error('Bun to run')
+		if (typeof Bun === 'undefined')
+			throw new Error(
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -3973,7 +3973,10 @@ export default class Elysia<
 		options: string | number | Partial<Serve>,
 		callback?: ListenCallback
 	) => {
-		if (!Bun) throw new Error('Bun to run')
+		if (typeof Bun === 'undefined')
+			throw new Error(
+				'.listen() is designed to run on Bun only. If you are running Elysia in other environment please use a dedicated plugin or export the handler via Elysia.fetch'
```
ThatOneCalculator, line 3978: "... designed to only run in Bun" sounds a bit better  instead of "...designed to run on Bun" imo

Issue #381, opened 2023-12-27
[BUG] Close function is passed wrong WebSocket


Kitanga:
The id of the websocket passed to `close` function isn't the same as the one sent to `open` function. Actually, according to my tests, it seems a new websocket is created with a new id.

If you run the following code

```javascript
.ws('/sockets', {
    open(ws) {
      console.log('Connection opened. ID:', ws.id)
    },
    close(ws, code, msg) {
      console.log('Connection Closed!!')
      console.log('code:', code)
      console.log('msg:', msg)
      console.log('Connection Closed ID:', ws.id)
      console.log('Date now:', Date.now())
    },
    message(ws) {
      // c
    }
  })
```

You'll get the following result

![image](https://github.com/elysiajs/elysia/assets/14119482/f18a2b26-bf96-4e33-815f-6b00bdf8db8c)

Notice how the `Date.now()` result is the same as the websocket id passed, I'm having a feeling the websocket is being recreated. ~~I'm also, currently checking if this is the same behaviour in vanilla Bunjs~~

Kitanga:
OK, so Bunjs' websocket implementation doesn't have IDs of any sort. But the websocket passed to the `open` and `close` function is the same. In Elysia, this isn't true, they are different. A work around would be to go through my list of connected sockets but even that doesn't seem plausible given that I can't see a way to check if the connection is still alive.

PR #383, opened 2023-12-27
Fix: Websocket ID not the same for `open` and `close` functions

Kitanga:
## Bug

This is to address this issue #381

When `open` and `close` functions are run, the passed `ElysiaWS` objects are different. This is due to the fact that each time `open`/`close` is run they recreate the `ElysiaWS` object which recreates the ID:

image

And here's the offending code `src/index.ts:3234` and `src/index.ts:3264`:

image

The fix I came to that doesn't break anything really is to use a getter/setter for the `id` prop on `ElysiaWS` class and storing the id in the raw bunjs/uWebSocket socket (`ElysiaWS.raw`).

## Moving away from `Date.now()`

You'll probably notice that I used `randomInt()` instead of `Date.now()`. This is because if you run `Date.now()`/`performance.now()` you'll notice that there are duplicates. Of course `performance.now()` has less duplicates but still for a simple test with only 100 iterations the best I could get was 97-98 none-duplicates.

I tried different solutions like `Date.now() + Math.random()` but still got duplicates. So `crypto`'s `randomInt` was the best I could find, had 100% success rate when creating unique IDs. But if you can think of a better more efficient solution, please do share.

Here's how I got to this solution, in case you end up trying to find a simpler answer:

- Just change the ID back to the correct one after creating `ElysiaWS` object in `close` function. (Problem was where does one store IDs and how does one get the ID back)
- Create `Map` with raw socket as key and ID as value. (whilst doing this I found the `data` prop in raw socket, ended up running with that instead)

### DIFF
--- a/example/websocket.ts
+++ b/example/websocket.ts
@@ -5,6 +5,10 @@ const app = new Elysia()
 	.ws('/ws', {
 		open(ws) {
 			ws.subscribe('asdf')
+			console.log('Open Connection:', ws.id)
+		},
+		close(ws) {
+			console.log('Closed Connection:', ws.id)
 		},
 		message(ws, message) {
 			ws.publish('asdf', message)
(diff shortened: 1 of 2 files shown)


DaviDecoder:
LGTM

SaltyAom (approved):

PR #413, opened 2024-01-12
fix: prevent information loss in string to number conversion

### LINKED ISSUE #410: cookie parsing may lose original value
cookie plugin tries to parse original raw string value as a better type: number, boolean, ...

I think cookie plugin should be able to provide raw original string value. I have a case where stored token may be `"0000000000000000000000000000000000000000"` which parses down to `0` or `1111111111111111111111111111111111111111` which numerically overflows.

ViliamKopecky:
Closes #410

Fixes cases when numeric string value should stay as string such as long tokens with numeric values like:

- `0000000000000000000000000000000000000000`
- `1111111111111111111111111111111111111111`

### DIFF
--- a/src/cookie.ts
+++ b/src/cookie.ts
@@ -2,7 +2,7 @@
 import { parse } from 'cookie'
 import type { Context } from './context'
 
-import { isNumericString, unsignCookie } from './utils'
+import { isParseableNumericString, unsignCookie } from './utils';
 import { InvalidCookieSignature } from './error'
 
 export interface CookieOptions {
@@ -431,7 +431,7 @@ export const parseCookie = async (
 			}
 
 		// @ts-ignore
-		if (isNumericString(value)) value = +value
+		if (isParseableNumericString(value)) value = +value
 		// @ts-ignore
 		else if (value === 'true') value = true
 		// @ts-ignore
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,7 +29,7 @@ import {
 	traceBackMacro,
 	replaceUrlPath,
 	primitiveHooks,
-	isNumericString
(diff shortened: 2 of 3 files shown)


fredericoo:
Looks good, I wonder if there isn’t a more performant way to do this?
E.g.: regexp
```/^[1-9]([\d.]{0,14}\d?)?$/```
this operates in the assumption of:
- the max integer (`562949953421311`) length is 16 digits
- the max float point with 2 decimal precision is `70368744177663` which has 14 digits + comma
- the only non-numerical digit is `.`
- numbers cannot end with `.`
- numbers cannot start with `.`

Im not even sure this would be faster by any means, and haven't benchmarked it either, just throwing in ideas

ViliamKopecky:
After some thought I believe there should be no implicit parsing - you should just get the string as is then you may parse it as you need.

fredericoo:
Found an easier and less intrusive solution to the problem:

The issue we're facing stems from the `isNumeric` function. It uses `Number.isNaN()` instead of `isNaN` which returns false for anything _but_ NaN, where we want it to return true to non-numbers too.

Found out because of (this thread)[https://twitter.com/slicknet/status/1750219872217522196] and it really checks out.

A simple, one liner fix

Chooks22:
This just bit me in the ass... right now I'm working with hex strings as session ids, and sometimes for unexpected reason the cookie's value would just turn up as `NaN`.

Workaround is to just prefix it with any non-number to force it to be parsed as string.

fredericoo:
Opened #446 as a simpler fix to the issue without requiring any new functions, and adding tests ◡̈

saketverma-03:
i have solved this issue in this PR #474

PR #422, opened 2024-01-15
fix(handler.ts): update switch case to check for contructor instead of constructor.name

### LINKED ISSUE #421: Returning a subclass of Response shows an empty object {}
Example code:

```ts
import { Elysia } from "elysia";

class ResponseTest extends Response {}

new Elysia()
    .all('/response', () =>{
        return new Response('Hi')
    })
    .all('/subclass', () =>{
        return new ResponseTest('Hi')
    })
    .listen(3000)

```
Result:
```
/response: Hi
/subclass: {}
```

Hi, based on this [docs](https://elysiajs.com/essential/handler.html#response), I should be able to return a Response class. It does work as shown in the code above. However, if I return a subclass of the Response, it returns an empty object.

This prevents me from creating a subclass of Response that contains different status code and headers. e.g. ResponseWithUnauthorized(). 

A workaround for this is to create a helper function that returns a Response class with the desired status code and header, but I prefer returning a subclass of Response :-)

hambergerpls:
This change tries to fix #421 and includes the following:

1. Refactor switch case logic for mapResponse functions to check against constructor function instead of constructor name. Also reordered the switch case to check for primitives first then the rest.

2. Use `instanceof` for `Response` and `Error` in `default` case to allow returning of subclasses of `Response` and `Error` as response

There is a [workaround](https://github.com/elysiajs/elysia/issues/421#issuecomment-1892045685), but it doesn't seem quite right.

Example code:
```ts
import { Elysia } from "elysia";

class ResponseTest extends Response {}

new Elysia()
    .all('/response', () =>{
        return new Response('Hi')
    })
    .all('/subclass', () =>{
        return new ResponseTest('Hi')
    })
    .listen(3000)
```

Result:
```
/response: Hi
/subclass: Hi
```

The reason for point (1) was that I was looking for some ways to switch case a type of a variable and found [this](https://stackoverflow.com/a/36332700). So I decided to give it a try and the tests are all green. One thing I noticed was that the tests finished a little bit quicker. So I benchmarked the before and after of this change.

Screenshot 2024-01-15 204958

From the benchmark, it seems that there is 11-14% improvement. I'm not sure if this is a good way to benchmark (if anyone can help me do some proper benchmark, feel free to help 🙏 ).

### DIFF
--- a/src/handler.ts
+++ b/src/handler.ts
@@ -154,18 +154,33 @@ export const mapResponse = (
 				set.headers['Set-Cookie']
 			) as any
 
-		switch (response?.constructor?.name) {
-			case 'String':
-				return new Response(response as string, set as SetResponse)
+		switch (response?.constructor) {
+			case undefined:
+				if (!response) return new Response('', set as SetResponse)
 
-			case 'Blob':
-				return handleFile(response as File | Blob, set)
-
-			case 'Object':
-			case 'Array':
 				return Response.json(response, set as SetResponse)
 
-			case 'ReadableStream':
+			case Boolean:
+			case Number:
+				return new Response(
+					(response as number | boolean).toString(),
+					set as SetResponse
+				)
+				
+			case String:
+				return new Response(response as string, set as SetResponse)
+
+			case Object:
+			case Array:
+				return Response.json(response, set as SetResponse)
+
+			case Function:
+				return mapResponse((response as Function)(), set)
+	
+			case Blob:
+				return handleFile(response as File | Blob, set)
+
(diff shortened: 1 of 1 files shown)


SaltyAom:
Thanks for your PR
Yes, #421 makes perfect sense integrating with 3rd Party Response APIs like NextResponse, ExpoResponse, and others.

The reason we are using constructor.name is because it's a more performance way to check for loose type, however, if non is matched I think we could implement the check using `instanceof`, so I'm going to modify your code for the next release, and close your PR.

Thank you.

PR #433, opened 2024-01-21
Remove crypto in favor of webcrypto

Mkassabov:
## What this does
Removes crypto api reference in favor of web crypto.
Should fix
- issues outlined in this thread: https://discord.com/channels/1044804142461362206/1194817027811053629
  - (vite failing when it tries to bundle elysia)
- cloudflare worker support

### DIFF
--- a/src/ws/index.ts
+++ b/src/ws/index.ts
@@ -7,7 +7,6 @@ import { ValidationError } from '../error'
 import type { Context } from '../context'
 
 import type { DecoratorBase, RouteSchema } from '../types'
-import { randomInt } from 'crypto'
 
 export const websocket: WebSocketHandler<any> = {
 	open(ws) {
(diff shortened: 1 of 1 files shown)


Mkassabov:
Also worth noting, this is the 3rd time I've seen an issue that seems to be related to importing crypto instead of using webcrypto.
Might be worth adding a test that just fails whenever crypto is imported to prevent these kinds of issues from happening

dtinth:
:+1:

This issue is preventing Elysia from working with Cloudflare workers. I added an issue on Cloudflare side here:

- https://github.com/cloudflare/workers-sdk/issues/4825

Even with `node_compat = true` the polyfill used by Wrangler is so old that it gives this error:

```
No matching export in "node-modules-polyfills:crypto" for import "randomInt"
```

PR #446, opened 2024-01-25
fix: numeric string check to use Number instead of parseInt

### LINKED ISSUE #410: cookie parsing may lose original value

fredericoo:
Closes #410 

Simple fix: numeric strings were being checked with `parseInt` which gobbles up non-numeric characters.

E.g.:
```ts
parseInt("9823a8")
// 9823

Number.isNaN(parseInt("9823a8"));
// false
```

by replacing it with `Number` we get realistic results:

```ts
Number("9823a8")
// NaN

Number.isNaN(Number("9823a8"))
// true
```
I’ve also added a few tests to make sure we don’t regress.


SaltyAom (approved):

SaltyAom:
Merged in local with commit [6439d40](https://github.com/elysiajs/elysia/commit/6439d40b4f3d58a148c70443d8d3aeaade8a1438) because of lockfile conflict.

Thanks for your contribution, closing.

PR #474, opened 2024-02-10
fix: Numeric Cookie with length >= 16 cant be parsed to number

### LINKED ISSUE #473: Bug: Cookie parser can't parse numeric value with length >= 16 digits properly

0.8.50

popos, linux x86_64

set a numeric cookie in browser with length greater then 16
ex: `ctx.cookie.id = 1234567890123456789`

when you use `ctx.cookie.id.get()` to read this youll get `1234567890123456700`

_No response_

### What do you see instead?

_No response_

## Cause of issue

https://github.com/elysiajs/elysia/blob/6fd8753fa2cf45f34a2af7f63a49659a987a87f7/src/cookie.ts#L434
 in this line we do `+value` to parse the number but js cant parse it properly

## Solution 
Numeric cookie value with length >= 16 should not be parsed

saketverma-03:
this fixes #473
in https://github.com/elysiajs/elysia/blob/6fd8753fa2cf45f34a2af7f63a49659a987a87f7/src/utils.ts#L586

added extra checks to  return false if `message.length > 16`
or
double check the string if  `message.length = 16` by converting to string and comparing to message

if less then 16 then do normal checking
```ts
	if (message.length < 16)
		return message.trim().length !== 0 && !Number.isNaN(Number(message))
```

```ts
	// if 16 digit but less then 9,007,199,254,740,991 then can be parsed
	if (message.length === 16) {
		const numVal = Number(message)
		if (numVal.toString() === message)
			return message.trim().length !== 0 && !Number.isNaN(numVal)
	}
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -583,4 +583,16 @@ export const traceBackMacro = (
 	}
 }
 
-export const isNumericString = (message: string) => message.trim().length !== 0 && !Number.isNaN(Number(message))
+export const isNumericString = (message: string): boolean => {
+	if (message.length < 16)
+		return message.trim().length !== 0 && !Number.isNaN(Number(message))
+
+	// if 16 digit but less then 9,007,199,254,740,991 then can be parsed
+	if (message.length === 16) {
+		const numVal = Number(message)
+		if (numVal.toString() === message)
+			return message.trim().length !== 0 && !Number.isNaN(numVal)
+	}
+
+	return false
+}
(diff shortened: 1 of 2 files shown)


bogeychan:
Can you add a test case for this please? Maybe under https://github.com/elysiajs/elysia/tree/main/test/cookie

saketverma-03:
> Can you add a test case for this please? Maybe under https://github.com/elysiajs/elysia/tree/main/test/cookie

i think i should add test in this file https://github.com/elysiajs/elysia/blob/main/test/units/numeric.test.ts
it allready contains some tests for `isNumericString()` util

saketverma-03:
@bogeychan i have added the test case, can you review them.

SaltyAom (approved):

Issue #246, opened 2023-10-01
Schema inheritance/merging between parent and child scopes


adicco:
Let's say you have a wishlist app and the wishlist objects all have items. You want standard CRUD routes on `/wishlist/:id` but also on `/wishlist/:id/items/:item_id`

I was trying to come up with a tidy way of reusing the "fetch wishlist" logic which includes not only doing the database call but also imposing some kind of type requirement on the params.

```ts
import Elysia, { t } from 'elysia';
const plugin = new Elysia()
  .guard({
    params: t.Object({
      id: t.String(),
    }),
  })
  .derive(async ({ params: { id } }) => {
    const wishlist = async fetchWishlistFromDb(id)
    return { wishlist };
  });
export default plugin;
```

I would then use this plugin in my route handlers:

```ts
const fetchWishlistPugin = require(the plugin above)
app
.use(fetchWishlistPugin)
.get('/wishlist/:id', async({ wishlist }) => { ... })
```

So far so good. But now let's say I want to implement the child routes for CRUD operations on the wishlist's items. I was doing something like:

```ts
const itemsRoutes = new Elysia()
.group('/wishlist/:id', (app) => {
      app
      .use(fetchWishlistPugin)
      .get('/items/:item_id', async({ wishlist, params: { item_id }) => {
         ...
      }, {
            params: t.Object({
              item_id: t.String()
            })
      })
});
```

I was hoping in my nested routes to have some kind of merge automagically happening on my `params` schema, but this raises an error that both `id` and `item_id` are defined while it is only expecting `item_id` as params.

```
Invalid params, 'id': Unexpected property

Expected: {
  "item_id": ""
}

Found: {
  "item_id": "xyb47ihmir7l97weu19ld88x",
  "id": "whatever"
}
```

Would there be any way of having some inheritance of the schema between parent and child scopes, and if not, how would one model such a case? It would feel neat to be able to encapsulate in a plugin the fetching of the wishlist along with its params' (or others, say body) requirements. Thanks!

adicco:
To fix my particular issue one can use typebox's ```additionalProperties```:

```ts
params: t.Object({
  item_id: t.String()
}, { additionalProperties: true })
```

But I am still wondering if there might be cases where merging/accessing the parent scope would be helpful?

davi-i:
[...]

PR #502, opened 2024-02-24
Merge response schema from parent scope

slippedandmissed:
Hello!

This relates to issue #246 and [issue 75 for elysia-swagger](https://github.com/elysiajs/elysia-swagger/issues/75)

## Summary:

If you have a parent scope (from a guard/plugin etc.) whose response schema is specified as a `Record<number, ...>`, and then you have a child scope which exposes another set of responses, e.g.

```typescript
const app = new Elysia()
  .guard({
    response: {
      401: t.String()
    }
  })
  .get(
    '/',
    () => { /* ... */ },
    {
      response: {
        200: t.String()
      }
    }
  )
```

You'd expect that the possible response codes the `/` route can return is `[401, 200]`. However, instead the child overrides the parent and the only possible response code is `200`. This causes issues for API documentation (i.e., using the swagger plugin), in which you might want to document that e.g., all endpoints which sit behind an `auth` guard have the potential to return a 401.

From what I understand, the same issue exists with `body`, `query`, etc., but a workaround exists because those fields are `TObject`s, and so are able to specify `{ additionalProperties: true }`, which results in a deep merge from the parent scope.

This PR contains a fix for this issue for the `response` field: If both parent and child specify `response` as a `Record<number, ...>` then they are shallow merged. If this behaviour is undesirable, the user can explicitly remove an inherited response code by specifying e.g. `401: undefined`.

Thanks!

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -124,6 +124,21 @@ export const primitiveHooks = [
 	'detail'
 ] as const
 
+function mergeResponses(a: InputSchema['response'], b: InputSchema['response']) {
+	
+	// If both are Record<number, ...> then merge them,
+	// giving preference to b.
+	type RecordNumber = Record<number, any>
+	const isRecordNumber = (x: typeof a | typeof b): x is RecordNumber =>
+		typeof x === 'object' && Object.keys(x).every(isNumericString)
+
+	if (isRecordNumber(a) && isRecordNumber(b)) {
+		return { ...(a as RecordNumber), ...(b as RecordNumber) }
+	}
+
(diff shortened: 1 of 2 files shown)


SaltyAom (approved):

SaltyAom:
Hi, thanks for your contribution.

Due to too many breaking changes from 0.8 to 1.0, I decided to push this code directly to skip merge conflict on da0fa83.

Your contribution will be mentioned in CHANGELOG under 1.0.1.

Thank you.

Issue #542, opened 2024-03-17
body.Check is not a function. (In 'body.Check(c.body)', 'body.Check' is undefined) in Elysia 1.0.0


tnfAngel:
### What version of Elysia.JS is running?

1.0.0

### What platform is your computer?

Linux 5.15.146.1-microsoft-standard-WSL2 x86_64 x86_64

### What steps can reproduce the bug?

setting type to 'json' and using t.Object for body sends this response: {"name":"TypeError","message":"body.Check is not a function. (In 'body.Check(c.body)', 'body.Check' is undefined)"}

Repro:
```ts
import { Elysia, t } from 'elysia';

new Elysia()
    .post(
        '/test',
        () => ({
            test: 'true'
        }),
        {
            type: 'json',
            body: t.Object({
                hi: t.String()
            }),
        }
    )
    .listen(8080);
```

Reported on discord: https://discord.com/channels/1044804142461362206/1044807567022493766/1218896455641862284

### What is the expected behavior?

Do not see this error

### What do you see instead?

 {"name":"TypeError","message":"body.Check is not a function. (In 'body.Check(c.body)', 'body.Check' is undefined)"}

### Additional information

Nothing is logged into the console. Is just the HTTP POST response.

kofta999:
It seems that the problem is because of name conflict in `body` variable, when I tried to log the handler function
```ts
const {
                handler,
                handleError,
                hooks: {
                        transform,
                        resolve,
                        beforeHandle,
                        afterHandle,
                        mapResponse: onMapResponse,
                        parse,
                        error: handleErrors,
                        onResponse
                },
                validator: {
                        body,
                        headers,
                        params,
                        query,
                        response,
                        cookie
                },
                utils: {
                        mapResponse,
                        mapCompactResponse,
                        mapEarlyResponse,
                        parseQuery,
                        isNotEmpty
                },
                error: {
                        NotFoundError,
                        ValidationError,
                        InternalServerError,
                        ParseError
                },
                schema,
                definitions,
                ERROR_CODE,
                getReporter,
                requestId,
                parseCookie,
                signCookie,
                decodeURIComponent,
                ELYSIA_RESPONSE
        } = hooks



        return async function handle(c) {





                c.headers = c.request.headers.toJSON()

c.cookie = await parseCookie(c.set, c.headers.cookie, undefined)
if(c.qi !== -1) {
[...]

PR #543, opened 2024-03-17
🔧 fix: parse json body when setting type to json

kofta999:
Fixes: #542 

Fixes a bug where an error would happen when trying to parse a body and the type is set to `type: "json"` or `type: "application/json"`

The problem was in composed handler where `body` variable in `try` block was used in checking instead of `body` validator.

Fixed by renaming the `body` variable in `try` block to `reqBody`

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -609,12 +609,12 @@ export const composeHandler = ({
 					case 'json':
 					case 'application/json':
 						if (hasErrorHandler)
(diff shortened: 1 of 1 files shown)


SaltyAom:
Hi, thank you for the fix.

Unfortunately, this was fixed in e2d2620, thanks for your contribution.

kofta999:
Np, working on it was fun enough

PR #546, opened 2024-03-17
fix: support for module promise in `Elysia#use`

aleclarson:
The following use case was broken: Calling `app.use(import("./routes"))` where the "./routes" module has a default export that is an Elysia instance.

I‘m doing `instanceof Elysia` checks on the promise result and the module‘s "default" export, so a more informative error can be thrown, where before the app would crash with some obscure TypeError with a message of "Right side of assignment cannot be destructured" within the `Elysia#_use` method.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2534,25 +2534,31 @@ export default class Elysia<
 				plugin
 					.then((plugin) => {
 						if (typeof plugin === 'function') {
-							return plugin(
-								this as unknown as any
-							) as unknown as Elysia
+							return plugin(this)
 						}
 
-						if (typeof plugin.default === 'function')
-							return plugin.default(
-								this as unknown as any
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -2534,25 +2534,31 @@ export default class Elysia<
...
 
-						if (typeof plugin.default === 'function')
-							return plugin.default(
-								this as unknown as any
-							) as unknown as Elysia
+						if (typeof plugin.default === 'function') {
+							return plugin.default(this)
+						}
 
-						return this._use(plugin as any)
```
aleclarson, line 2547: The issue was that `plugin.default` was not being referenced here, i.e. `plugin.default || plugin`. But also, I decided to clean up this area of the code, removing unnecessary type casts and adding `instanceof` checks for proper type narrowing.

Issue #358, opened 2023-12-14
Response type auto stripping not declared fields or disable type checking?


m1212e:
I use the response field for documentation purposes so the generated swagger output has the correct return types set. But sometime I return objects which satisfy the response TS type but contain additional fields which I got e.g. from a more extensive DB query. Now when I return these, TS doesn't complain, since it is a sufficient type, but at runtime the request fails, since the returned object has an additional field.

Is there some way to disable these kinds of typechecks at runtime, or maybe even better, automatically strip all additional fields automatically to comply to my return schema?

m1212e:
Maybe something like this would be an elegant way? Did you consider bringing this into elysia?

https://github.com/sinclairzx81/typebox/issues/278

kravetsone:
[Elysia normalize response by default since 1.1](https://elysiajs.com/blog/elysia-11.html#normalization)

This issue can be closed

PR #562, opened 2024-03-20
✨ feat: add option for body cleaning

m1212e:
This adds the option to simply clear the incoming and outgoing bodies instead of returning/raising an error. Useful for usage with strongly typed ORMs like prisma (https://github.com/elysiajs/elysia/issues/358). This works both ways, also incoming requests don't have to worry about getting rejected when providing sufficient data, it will simply be stripped and put into the expected schema so there is no way of leaking unexpected data into the handler.


m1212e:
Thanks for merging 🥳 <3

PR #593, opened 2024-04-05
Fix 204 returning internal server error

kevainc-plop:
When a route with a 204 would be called it would cause an internal server error with the message: 

> undefined is not an object (evaluating 'r[ELYSIA_RESPONSE]'

```
new Elysia()
.get("/fails", ({ set }) => {
  set.status = 204;
}, {
    response: {
      204: t.Void()
    }
})
```
This PR fixes this and added a test to avoid regression.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -157,7 +157,7 @@ const composeValidationFactory = (
 
 		let code = '\n' + injectResponse + '\n'
 
-		code += `const er = ${name}[ELYSIA_RESPONSE]\n`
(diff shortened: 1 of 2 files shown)


SaltyAom:
Thanks for your contribution.

Unfortunately, we have shipped the fix in this commit, https://github.com/elysiajs/elysia/commit/0b2b04d631d6aecb71effea8f8bee2de62aba754.

Thank you for work nevertheless 😊

PR #601, opened 2024-04-10
chore: fix typo in compose.ts

mtt-artis:


SaltyAom:
Could you explain why you propose this change?
Is there any incorrect behavior that lead to this change?

mtt-artis:
Hello, 
no particular error, 
just a line that repeats the one above.

Thanks for maintaining this awesome project!

SaltyAom:
Ah yes, you're right, I somehow misread it as onRequest and onResponse haha.
Thanks.

PR #636, opened 2024-05-13
Fix #635 - Use StaticDecode<> instead of Static<> to get type of inpu…

### LINKED ISSUE #635: Type inference of `query` params is wrong when Transforms are used

elysia@1.0.18

Linux 4.4.0-19041-Microsoft x86_64 x86_64

Create a type transform like:

```ts
import {t} from "elysia"

export const PageToken = () => t.Transform(t.String())
    .Decode(decodeToken)
    .Encode(encodeToken)
```

Then type your route:

```ts
{
        query: t.Object({
            afterCursor: PageToken(),
        })
    }
```

`query.afterCursor` should have the type of the return value of `decodeToken`

### What do you see instead?

`query.afterCursor` is typed as `string` (the *input* type coming from the client)

### Additional information

I haven't looked at Elysia's source code yet, but I'm guessing Elysia us using `Static` to infer the type when `StaticDecode` should be used instead: https://github.com/sinclairzx81/typebox?tab=readme-ov-file#transform-types

There's another example of transforms on that page that's perhaps better than mine.

Edit: I think this line is the culprit:

https://github.com/elysiajs/elysia/blob/968650becd19850e4f33dca273d9e5838c937d70/src/types.ts#L205

mnpenner:
…t schema

This will infer the correct type with Typebox Transform Types (https://github.com/sinclairzx81/typebox?tab=readme-ov-file#transform-types) are used

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,7 +2,7 @@
 import type { Elysia } from '.'
 import type { Serve, Server, WebSocketHandler } from 'bun'
 
-import type { TSchema, TObject, Static, TAnySchema } from '@sinclair/typebox'
+import type { TSchema, TObject, StaticDecode, TAnySchema } from '@sinclair/typebox'
 import type { TypeCheck } from '@sinclair/typebox/compiler'
 
 import type { OpenAPIV3 } from 'openapi-types'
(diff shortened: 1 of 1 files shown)


david-plugge:
@SaltyAom do you consider this for the next release? This is pretty useful. Like transforming a string to a date, converting using an enum and so on. I wonder if this is actually just a type issue or if this needs more work (actually transforming the data using typebox).

mnpenner:
> I wonder if this is actually just a type issue or if this needs more work (actually transforming the data using typebox).

I didn't hunt down where the deserialization happens, but as long as Elysia is using [`Decode`](https://github.com/sinclairzx81/typebox?tab=readme-ov-file#decode) it should automagically work, and indeed that seems to be the case in my little project.

SaltyAom (approved):

Issue #631, opened 2024-05-10
Normalize disabled when additional properties is set to true


m1212e:
Hi @SaltyAom,

at 
/src/utils.ts:273
/src/utils.ts:347

you check the following condition:

```ts
if (normalize && schema.additionalProperties === true)
			// @ts-ignore
			compiledValidator.Clean = cleaner
```
which I think should be the other way around: If normalization is enabled and the schema DOES NOT allow additionalProperties the cleaner should kick in, since then we need to clean out the disallowed fields. If I understand the code correctly, currently we check if normalization is enabled, and if so we only clean when the schema allows additionalProperties, which doesn't make sense since we actually want the additional properties to pass?

PR #654, opened 2024-05-22
🐞 fix: set correct normalization behavior for addtional properties

m1212e:
Fixes https://github.com/elysiajs/elysia/issues/631 by adjusting the behavior for normalization with additional properties settings to a more expectable way. Uses sensible defaults.


m1212e:
Hey, @SaltyAom could you take a quick look at this? Would help me a lot! Thanks for you work!

SaltyAom (approved):

PR #657, opened 2024-05-27
Make "isProduction" default to true unless explicitly in development

7f8ddd:
This causes, for example, /bun:info to leak system information, when the NODE_ENV is not explicitly set to production. Some users may use prod, dev, staging, etc., so this fixes that problem.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -16,7 +16,7 @@ export const ERROR_CODE = Symbol('ElysiaErrorCode')
 export const ELYSIA_RESPONSE = Symbol('ElysiaResponse')
 export type ELYSIA_RESPONSE = typeof ELYSIA_RESPONSE
 
-export const isProduction = (env?.NODE_ENV ?? env?.ENV) === 'production'
+export const isProduction = (env?.NODE_ENV ?? env?.ENV) !== 'development'
 
 export type ElysiaErrors =
 	| InternalServerError


butcher73:
Defaulting to production can hide important debugging information, increase the risk of accidental deployments and security issues, and complicate the development process, so it's safer to default to development mode.

kravetsone:
> This causes, for example, /bun:info to leak system information, when the NODE_ENV is not explicitly set to production. Some users may use prod, dev, staging, etc., so this fixes that problem.

bad idea

7f8ddd:
> Defaulting to production can hide important debugging information, increase the risk of accidental deployments and security issues, and complicate the development process, so it's safer to default to development mode.

Except this only applies to the Bun server and has no effect on what environment variable you're using. If you're using this export, you have other problems.

butcher73:
I see, I misunderstood you at first. However, I still don't agree with setting it to 'prd' by default. I believe it's better to require the environment variable to be explicitly set before running. This way, the user must intentionally specify the mode, ensuring they have considered whether they want to run in development or production mode.

7f8ddd:
> I see, I misunderstood you at first. However, I still don't agree with setting it to 'prd' by default. I believe it's better to require the environment variable to be explicitly set before running. This way, the user must intentionally specify the mode, ensuring they have considered whether they want to run in development or production mode.

Why would the user want the internal Bun server in development mode? All it does is leak information. If anything, it should be made a config option.

PR #691, opened 2024-06-24
feat: add schema options to Nullable and MaybeEmpty.

MAST1999:


MAST1999:
After checking through the swagger plugin's code, I found that if you add the extra properties it'll show up, so this is not needed.

PR #714, opened 2024-07-13
Port can be dynamically configured

sunshinego12138:
decorate is read-only.so It's best for dynamically configuring ports
image

### DIFF
--- a/example/port.ts
+++ b/example/port.ts
@@ -1,0 +1,13 @@
+import { Elysia } from '../src'
+
+new Elysia()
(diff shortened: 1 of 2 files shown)


struckchure:
This looks good, but it feels ambiguous actually, the `({ port }) => port` part.
Log message part, I can relate to, although, it can be done differently, by leaving it as it is.

If we all stand by, "If it's not broken, don't fix it", let's keep this part simple.

We love Elysia ❤️ !

PR #718, opened 2024-07-15
🐞 fix: implement normalization support for class instances with getter functions

m1212e:
When returning class instances which use getter functions, normalization does not work in stripping those fields. Typeboxes `Clean` does not remove getter instances on objects in certain scenarios. This came to my mind because the prisma ORM uses this when defining extensions resulting in the inability to return database query outputs directly. My PR adjusts the clean call to traverse the prototype tree of a passed object, gathering all values located behind a get call and merging them with the original value. It does so in a performance sensitive manner, I check if the operations are required before blindly applying them.

Would be great if you could take a look @SaltyAom and I'd be happy to receive feedback/make any necessary changes. Thanks for your work!


m1212e:
https://github.com/sinclairzx81/typebox/issues/842

PR #730, opened 2024-07-17
Fix package issues

kravetsone:
When I was using elysia and went into its `package.json` thanks to [`vscode-publint`](https://github.com/kravetsone/vscode-publint), I saw errors in the package configuration:

```bash
PS Z:\PROJECTS\forks\elysia> bunx publint
elysia lint results:
Suggestions:
1. The package does not specify the type field. Environments may incorrectly identify a CJS file as ESM in the future. Consider adding "type": "commonjs".
Warnings:
1. pkg.exports["."].types types is interpreted as CJS when resolving with the "import" condition. This causes the types to be ambiguous when default importing the package due to its implied interop. Consider splitting out two "types" conditions for "import" and "require", and use the .mts extension, e.g. pkg.exports["."].import.types: "./dist/index.d.mts"   
2. pkg.exports["."].bun is ./dist/bun/index.js and is written in ESM, but is interpreted as CJS. Consider using the .mjs extension, e.g. ./dist/bun/index.mjs
Errors:
1. pkg.exports["."].types should be the first in the object as required by TypeScript.
2. pkg.exports["./cookie"].types is ./dist/cookie.d.ts but the file does not exist.
3. pkg.exports["./cookie"].import is ./dist/cookie.mjs but the file does not exist.
4. pkg.exports["./cookie"].require is ./dist/cjs/cookie.js but the file does not exist.
5. pkg.exports["./schema"].types is ./dist/schema.d.ts but the file does not exist.
6. pkg.exports["./schema"].import is ./dist/schema.mjs but the file does not exist.
7. pkg.exports["./schema"].require is ./dist/cjs/schema.js but the file does not exist.
```

after my Pull Request

```bash
PS Z:\PROJECTS\forks\elysia> bunx publint
elysia lint results:
Warnings:
1. pkg.exports["."].types types is interpreted as CJS when resolving with the "import" condition. This causes the types to be ambiguous when default importing the package due to its implied interop. Consider splitting out two "types" conditions for "import" and "require", and use the .mts extension, e.g. pkg.exports["."].import.types: "./dist/index.d.mts"   
2. pkg.exports["."].bun is ./dist/bun/index.js and is written in ESM, but is interpreted as CJS. Consider using the .mjs extension, e.g. ./dist/bun/index.mjs
```

change `cookie` export to `cookies` and `schema` to `sucrose` (because he is the only one without exports)

also added `bunx publint` to the `test` script to prevent it cause again

and maybe we can drop CommonJS build? 

https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

> Since Node.js 22, you may be able to require() ESM modules. However, I strongly recommend moving to ESM instead.

It is also not needed in `Bun`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
 	"name": "elysia",
 	"description": "Ergonomic Framework for Human",
+	"type": "commonjs",
 	"version": "1.1.2",
 	"author": {
 		"name": "saltyAom",
@@ -13,8 +14,8 @@
 	"exports": {
 		"./package.json": "./package.json",
 		".": {
-			"bun": "./dist/bun/index.js",
 			"types": "./dist/index.d.ts",
+			"bun": "./dist/bun/index.js",
 			"import": "./dist/index.mjs",
 			"require": "./dist/cjs/index.js"
 		},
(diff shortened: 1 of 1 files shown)


kravetsone:
Also switch from rimraf to rm -rf because Bun shell is great work

Issue #741, opened 2024-07-20
Async generator yielding an object results in `[object Object]` output


remorses:
### What version of Elysia.JS is running?

"elysia": "^1.1.3"

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

```ts
import { treaty } from '@elysiajs/eden'
import { sleep } from 'bun'
import { Elysia } from 'elysia'

const app = new Elysia()
    .get('/', async function* generator() {
        await sleep(100) // needed because of another bug
        yield { body: 'Hello Elysia' } // this will be [object Object]
    })
    .listen(3000, async () => {
        const app = treaty<App>('localhost:3000')

        {
            const { data: gen, error } = await app.index.get({})
            if (error) throw error

            for await (const res of gen) {
                console.log(res)
            }
        }
    })

export type App = typeof app

console.log(
    `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`,
)
```

### What is the expected behavior?

Elysia should call JSON.stringify on every output of an async generator

### What do you see instead?

Elysia encodes objects to `[object Object]`

SaltyAom:
Should have been fixed with https://github.com/elysiajs/elysia/commit/0a4977acd6459c3357ad1dbe5d9231f7c7cd6278, published under 1.1.4

remorses:
Can you check https://github.com/elysiajs/elysia/pull/743? The server should return valid text/event-stream content instead of concatenated JSON strings

remorses:
For generators to make sense you need to have a separator between them, otherwise you can't distinguish each value. `text/event-stream` already has a format for this

remorses:
Closing in favor of #742

PR #743, opened 2024-07-20
Fix generators, use `text/event-stream` protocol

### LINKED ISSUE #742: Async generator handler should add a new line between each yielded item and call JSON.stringify on items to follow text/event-stream

"elysia": "^1.1.3"

Darwin 23.5.0 arm64 arm

```ts
import { treaty } from '@elysiajs/eden'

const app = new Elysia()
    .get('/', async function* generator() {
        await sleep(100)
        yield JSON.stringify({ body: 'Hello Elysia' })
        yield JSON.stringify({ body: 'Hello Elysia' })
        yield JSON.stringify({ body: 'Hello Elysia' })
        yield JSON.stringify({ body: 'Hello Elysia' })
        yield JSON.stringify({ body: 'Hello Elysia' })
    })
    .listen(3000, async () => {
        const app = treaty<App>('localhost:3000')

        {
            const { data: gen, error } = await app.index.get({})
            if (error) throw error

            for await (const res of gen) {
                console.log(res)
            }
        }
    })

export type App = typeof app

console.log(
    `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`,
)

function sleep(ms: number) {
    return new Promise(resolve => setTimeout(resolve, ms))
}
```

The output of this code will be:

```
{"body":"Hello Elysia"}{"body":"Hello Elysia"}{"body":"Hello Elysia"}
{
  body: "Hello Elysia",
}
{
  body: "Hello Elysia",
}
```

There are several problems with this code:
1. In Eden some items will be objects, sometimes strings. This is because Elysia cannot distinguish what separate each yieled item. This is because the server does not separate yielded items with new lines (as the text/event-stream spec)
2. The type inference of Eden is wrong because when you yield a string it will be converted to an object. The server should call JSON.stringify on all yielded items if the handler is an async generator. (also this is what the text/event-stream spec says)
3. Elysia just calls .toString() on the yielded items, this will cause problems where when you yield objects you will get `[object Object]` #741 

https://github.com/elysiajs/elysia/blob/8173d905944b8b305f8989985abe7760af681de9/src/handler.ts#L156

Elysia should return a `text/event-stream` compatible output if the handler is an async generator:
1. Separate each yielded object with a new line
2. JSON.stringify each item

### What do you see instead?

Elysia concatenates the outputs inline calling .toString() on each of them

### Additional information

_No response_

remorses:
Fix #742 
Fix https://github.com/elysiajs/elysia/issues/741


remorses:
What do you think @SaltyAom is this a breaking change or a bug fix? Previously the response was not following the `text/event-stream` protocol

luismanfroni:
Don't know if it's related, but having a better way to catch Aborts would be really great, instead of having to do something like 
`request.signal.addEventListener("abort", cb)` (which is kinda ugly) inside all request generators handlers for cleanups.
Would be cool an "abort" hook, or something like that.

remorses:
@SaltyAom do you need any help with the review? 

I am already using this fork in production and it works great. Another problem of the current approach is that most servers won't flush the body until they find a new line, using the right [`text/event-stream`](https://html.spec.whatwg.org/multipage/iana.html#text/event-stream) protocol fixes that

src/handler.ts
```diff
@@ -117,18 +117,24 @@ export const serializeCookie = (cookies: Context['set']['cookie']) => {
...
 	set?: Context['set'],
 	request?: Request
 ) => {
-	let init = generator.next()
-	if (init instanceof Promise) init = await init
-
-	if (init.done) {
-		if (set) return mapResponse(init.value, set, request)
-		return mapCompactResponse(init.value, request)
+	let init
```
SaltyAom: You can remove try catch in here, `handleStream` function is already wrapped in try-catch that handle onError already

src/handler.ts
```diff
@@ -146,38 +152,29 @@ const handleStream = async (
 					}
 				})
 
-				if (init.value !== undefined && init.value !== null) {
-					if (
-						typeof init.value === "object"
+				if (init?.value !== undefined && init?.value !== null)
+					controller.enqueue(
+						Buffer.from(
+							`event: message\ndata: ${JSON.stringify(init.value)}\n\n`
```
SaltyAom, line 152: Would make sense to check typeof object before using JSON.stringify

src/handler.ts
```diff
@@ -146,38 +152,29 @@ const handleStream = async (
...
+						if (end) break
+						if (chunk === undefined || chunk === null) continue
 
-					if (typeof chunk === 'object')
-						try {
-							controller.enqueue(
-								Buffer.from(JSON.stringify(chunk))
+						controller.enqueue(
+							Buffer.from(
+								`event: message\ndata: ${JSON.stringify(chunk)}\n\n`
```
SaltyAom, line 163: Same as above (check typeof of object before passing it to JSON.stringify)

SaltyAom (changes requested):
Hey, thanks a lot for your PR.
Would be great if you can help me with these changes and https://github.com/elysiajs/eden/pull/114, after that let's get this merge.

src/handler.ts
remorses: Currently the app will crash if you throw an error in a generator before yield, i added a test that reproduces this here: https://github.com/remorses/elysia/blob/932951679c89aef6bad08779b6eb7fe94c30335d/test/response/stream.test.ts#L75

src/handler.ts
remorses: If you don't always call JSON.stringify and the user yields a string that contains 2 new lines the response will break the text/event-stream protocol. If you always call JSON.stringify you can be sure that all the new lines are encoded with \n and put in a single line.

The parsing in Eden also becomes simpler because you know messages are always a JSON encoded string instead of using heuristics.

src/handler.ts
remorses: I fixed this crash removing the try catch when aot is false. In aot mode it still crashes.

the app crashes because the error is throws asynchronously while mapResponse is a not async, i think there are many more of these bugs hidden in the code, the ideal solution would be to always await `mapResonse` in aot too

HK-SHAO:
@SaltyAom @remorses In fact I don't think it should be up to elysia stream to handle exactly how to send the streams of the sse interface, that should be left to the developer. It should be left to the developer, because exactly how to send each id, event, etc. should be set by the developer.

remorses:
@HK-SHAO if you want customization you will need to implement it from scratch

```

app.get('/events', async ({ set }) => {
  set.headers['Content-Type'] = 'text/event-stream'
  set.headers['Cache-Control'] = 'no-cache'

  const stream = new ReadableStream({
    async start(controller) {
      for (let count = 0; count < 10; count++) {
        const event = `data: Event ${count}\n\n`
        controller.enqueue(event)
        await Bun.sleep(1000)
      }
      controller.close()
    }
  })

  return new Response(stream)
})
```

kravetsone:
> @HK-SHAO if you want customization you will need to implement it from scratch
> 
> ```ts
> app.get('/events', async ({ set }) => {
>   set.headers['Content-Type'] = 'text/event-stream'
>   set.headers['Cache-Control'] = 'no-cache'
> 
>   const stream = new ReadableStream({
>     async start(controller) {
>       for (let count = 0; count < 10; count++) {
>         const event = `data: Event ${count}\n\n`
>         controller.enqueue(event)
>         await Bun.sleep(1000)
>       }
>       controller.close()
>     }
>   })
> 
>   return new Response(stream)
> })
> ```

Yeah i agree. Elysia should folows specs but if you need sometning custom you can do it

remorses:
I created a full framework where async generators have first class support and use SSE, it's called [Spiceflow](https://github.com/remorses/spiceflow)

SaltyAom:
I think I have to added input to this for the record

> The problem is there are both use case for both text/event-stream and non text/event-stream
eg. stream a response in which response is not a text eg. image, video or create a proxy from AI provider like OpenAI, Claude, etc which are using direct streaming not text/event-stream, or create a proxy from fetch (Elysia can `return fetch(url)` and it will just become a proxy which is possible because of direct streaming)
I personally haven't come up with an idiomatic API yet which is why I leave it as it until then because I'm afraid of introducing a breaking changes.

Issue #412, opened 2024-01-12
swagger : hide endpoint from swagger


primadi:
Is it possible to hide specific endpoint from swagger doc ?

thanks.

kravetsone:
You can use exclude Regex)
```ts
.use(swagger({
            exclude: /^\/(auth|github)/i //regexp

}));
```


@SaltyAom 
in @fastify/swagger we can hide the route in the `schema` property
It would be more convenient if we could `hide: true` to `datails` to exclude the route from swagger

```ts
.get("/", "Hello, Elysia!", {
    detail: {
        hide: true
    }
});
```

kravetsone:
@SaltyAom @primadi i guess we can close it because https://github.com/elysiajs/elysia/pull/763 was merged

PR #763, opened 2024-08-02
feat: add `hide` props in `detail` types for hide route from OpenAPI/swagger document

kravetsone:
Exclude from OpenAPI/swagger file routes with inspired by [@fastify/swagger hide paths](https://github.com/fastify/fastify-swagger?tab=readme-ov-file#hide-a-route)

```ts
detail: {
    hide: true
}
```

example

```ts
const app = new Elysia()
                        .use(swagger())
			.get("/public", "omg")
			.guard({
				detail: {
					hide: true
				}
			})
			.get("/hidden", "ok")
```

Should be merged with https://github.com/elysiajs/elysia-swagger/pull/137

Close/helps https://github.com/elysiajs/elysia/issues/412

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -940,7 +940,12 @@ export type Isolate<T> = {
 	[P in keyof T]: T[P]
 }
 
-export type DocumentDecoration = Partial<OpenAPIV3.OperationObject>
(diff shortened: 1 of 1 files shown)


zoto-ff:
@SaltyAom

PR #839, opened 2024-09-20
Fix `INVALID_COOKIE_SIGNATURE` error not calling `onError` handler

### LINKED ISSUE #707: onError handler isn't getting called during INVALID_COOKIE_SIGNATURE

1.0.26

Microsoft Windows NT 10.0.22631.0 x64

```ts
 .onError(({ code, set, cookie: { accessToken, refreshToken } }) => {
    if (code === "NOT_FOUND")
      return { error: "Route not found :(", status: STATUS.ERROR };

    if (code === "INVALID_COOKIE_SIGNATURE")
      return { error: "Your cookies has been altered", status: STATUS.ERROR };

    set.status = 500;
    return { error: "Internal Server Error", status: STATUS.ERROR };
  })
```

When cookies have been altered, it should respond with the format `{ error: "...", status: "ERROR" }` 

### What do you see instead?

I am just getting raw text when cookies have been altered. onError is not getting called.

### Additional information

_No response_

nxht:
- fixes #707
- test added under `test/lifecycle/error.test.ts`
   - Not sure if it's the correct place to be instead of `core/handler-error.test.ts` or `cookie/signature-test.ts`
- Simply moved the `hasCookie` after `try {` (before body parsing)
  - It might be better to give separate try catch but I wasn't sure 
  - It made more sense to be move the logic after body parsing like in dynamic handler case but I was trying to minimize the change.
  - Also, this leaves other errors like `t.Cookie required secret which is not set in ...` to still have same issues

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -565,65 +565,6 @@ export const composeHandler = ({
 				`
 	}
 
-	if (hasCookie) {
-		const get = (name: keyof CookieOptions, defaultValue?: unknown) => {
-			// @ts-ignore
-			const value = cookieMeta?.[name] ?? defaultValue
-			if (!value)
-				return typeof defaultValue === 'string'
-					? `${name}: "${defaultValue}",`
-					: `${name}: ${defaultValue},`
-
-			if (typeof value === 'string') return `${name}: '${value}',`
-			if (value instanceof Date)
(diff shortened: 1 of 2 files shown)


SaltyAom:
added test case via d4a4c6b

somehow the test case passed, so I assumed it was fixed somewhere in between

PR #863, opened 2024-10-04
Fix: Typebox isn't able to deref nested references

aberigle:
`references` weren't being passed along to `Clean` when validating a response. So Typebox wasn't able to deref.

I've added the references in the cleaner, to be able to deref properly.

There is a new test case as an example also:

```typescript
  it('validate nested references', async () => {
    const job = t.Object({
      name : t.String()
    }, { $id: "job" })

    const person = t.Object({
      name : t.String(),
      job  : t.Ref(job)
    })

    const app = new Elysia()
    .model({ job, person })
    .get('/', () => ({
      name : 'sucrose',
      job: { name: 'alchemist' }
    }), {
      response : person
    })

    const res = await app.handle(req('/'))
    expect(res.status).toBe(200)
  })
```

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -714,10 +714,10 @@ export const getResponseSchemaValidator = (
 	const compile = (schema: TSchema, references?: TSchema[]) => {
 		const cleaner = (value: unknown) => {
 			if (!value || typeof value !== 'object')
-				return Value.Clean(schema, value)
+        return Value.Clean(schema, references || [], value)
 
-			if (Array.isArray(value)) value = Value.Clean(schema, value)
-			else value = Value.Clean(schema, value)
+      if (Array.isArray(value)) value = Value.Clean(schema, references || [], value)
+      else value = Value.Clean(schema, references || [], value)
 
 			return value
 		}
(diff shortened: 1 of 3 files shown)


aberigle:
Hi @SaltyAom, sorry to tag you :sweat_smile: , but any thoughts on this?

src/utils.ts
```diff
@@ -714,10 +715,10 @@ export const getResponseSchemaValidator = (
 	const compile = (schema: TSchema, references?: TSchema[]) => {
 		const cleaner = (value: unknown) => {
 			if (!value || typeof value !== 'object')
-				return Value.Clean(schema, value)
+        return Value.Clean(schema, references || [], value)
 
-			if (Array.isArray(value)) value = Value.Clean(schema, value)
-			else value = Value.Clean(schema, value)
+      if (Array.isArray(value)) value = Value.Clean(schema, references || [], value)
+      else value = Value.Clean(schema, references || [], value)
```
crishoj: Would suggest adhering to @SaltyAom's convention of using tabs for indentation 🙂

SaltyAom:
added test case by 701e3de

somehow the test case passed, so I assume that it got fixed somewhere in between

PR #889, opened 2024-10-21
:wrench: fix: handle `error()` in resolve with `aot: false`

### LINKED ISSUE #888: Cannot return error on resolve when set ElysiaConfig aot to false

"elysia": "1.1.22"

Darwin 24.0.0 arm64 arm

```
import Elysia, { error } from "elysia";

const app = new Elysia({ aot: false })
  .resolve(() => {
    return error(401);
  })
  .get("/hello", () => "Hello")
  .listen(3000);
```

When we call endpoint `/hello` should be response an error status code `401`.

### What do you see instead?

_No response_

### Additional information

_No response_

yes

crishoj:
Fixes #888

@SaltyAom — it seems a little inconsistent to sprinkle cases with `{ aot: false }` over the tests. Perhaps you have a more exhaustive approach for full test coverage both with and without AOT?

### DIFF
--- a/src/dynamic-handle.ts
+++ b/src/dynamic-handle.ts
@@ -294,6 +294,10 @@ export const createDynamicHandler =
 				let response = hook.fn(context)
 
 				if (hook.subType === 'resolve') {
+					if (response instanceof ElysiaCustomStatusResponse) {
+						const result = mapEarlyResponse(response, context.set)
+						if (result) return (context.response = result)
+					}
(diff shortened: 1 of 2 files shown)


crishoj:
@SaltyAom more generally, I think Elysia maintainers should consider either
- following @remorses's approach in #773 of full test coverage for both `{aot: true}` and `{aot: false}`, or
- deprecating/disabling `{aot: false}`.

PR #891, opened 2024-10-23
🔧 fix: Upgrade Cookie to 0.7.x, fix CVE-2024-47764

nathankleyn:
Cookie `0.6.x` is vulnerable to an XSS vulnerability which has been logged as CVE-2024-47764 at
https://www.cve.org/CVERecord?id=CVE-2024-47764. This commit upgrades cookie to `^0.7.0` which includes the CVE fix. There is cookie `1.0.x` available now but it includes breaking changes which should be done separately.

There is no currently available upgrade to `@types/cookie` for `0.7.x` but `0.6.x` is type compatible.


kravetsone:
yeah we should migrate to latest soon

Sly777:
@SaltyAom is there any planned date for this? It's security issue on elysia at the moment.

PR #905, opened 2024-10-31
Avoid response normalization side-effects

### LINKED ISSUE #897: `normalize: true` mutates objects used in responses

1.1.23

macOS

With [normalize](https://elysiajs.com/patterns/configuration.html#normalize) `true` (the default), Elysia will _mutate_ objects used in responses.

Consider this example with a long-lived object `people`:

```ts
import {Elysia, t} from 'elysia'

const people = new Map<number, object>([
    [1, {name: 'Jane Doe', age: 80}],
])

const app = new Elysia()
    .get('/names/:id', ({params: {id}}) => {
        return people.get(id)
    }, {
        params: t.Object({id: t.Number()}),
        response: t.Object({name: t.String()})
    })

console.log('people before request:', people)
await app.handle(new Request('http://localhost/names/1'))
console.log('people after request:', people)
```

```
people after request: Map(1) {
  1: {
    name: "Jane Doe",
    age: 80,
  },
}
```

From a developer perspective, I expect Elysia to not directly mutate objects used in a response body.

If response normalization is active, I expect cleaning to happen on a (deep) clone of the object passed as response body.

### What do you see instead?

```
people after request: Map(1) {
  1: {
    name: "Jane Doe",
    // ⚠️ Notice that the `age` property is now missing from the long-lived object
  },
}
```

Objects used in responses are directly mutated — additional properties are deleted by the Typebox cleaner:

```ts
if (IsSchema(additionalProperties) && Check(additionalProperties, references, value[key])) {
    value[key] = Visit(additionalProperties, references, value[key]);
    continue;
}
delete value[key];
```

### Additional information

This caught me off guard and led to subsequent errors in my app where properties in a long-lived object were suddenly undefined.

For improved DX, please consider doing a `structuredClone()` of objects before normalization.

Yes

crishoj:
The Typebox cleaner _mutates_ the cleaned value:

> Clean is a mutable operation. To avoid mutation, Clone the value first.

```ts
if (IsSchema(additionalProperties) && Check(additionalProperties, references, value[key])) {
    value[key] = Visit(additionalProperties, references, value[key]);
    continue;
}
delete value[key];
```

Therefore, it can break type-safety and lead to subtle bugs (and poor DX) when long-lived server objects are used in responses.

This proposed fix here is to use `structuredClone()` on the response body _before_ cleaning.

Fixes #897


src/utils.ts
```diff
@@ -715,13 +715,7 @@ export const getResponseSchemaValidator = (
...
 	const compile = (schema: TSchema, references?: TSchema[]) => {
 		const cleaner = (value: unknown) => {
-			if (!value || typeof value !== 'object')
-				return Value.Clean(schema, value)
-
-			if (Array.isArray(value)) value = Value.Clean(schema, value)
-			else value = Value.Clean(schema, value)
-
-			return value
+			return Value.Clean(schema, structuredClone(value))
```
crishoj, line 718: @SaltyAom correct me if I'm wrong, but it seems to me all 3 logical branches end up doing the same thing?

SaltyAom (approved):

crishoj:
🙏

PR #953, opened 2024-12-20
feat: support `t.Integer()`

### LINKED ISSUE #949: `t.Integer()` doesn't work with query params

1.1.26

WSL Ubuntu

```ts
// test.test.ts
import { test, expect } from "bun:test";
import { Elysia, t } from "elysia";

test("Integer query param", async () => {
  const app = new Elysia().get("/", ({ query: { limit } }) => limit, {
    query: t.Object({
      limit: t.Integer(),
    }),
  });

  await app.modules;

  const response = await app.handle(new Request(`http://localhost/?limit=1`));

});
```

`bun test`

test pass

### What do you see instead?

test fails:

```
Expected: 200
Received: 422
```

### Additional information

Should work out of the box, like `t.Number()`. We can't use `t.Number()` here because it allows floating point numbers.

---

reported on [discord](https://discord.com/channels/1044804142461362206/1044804143455420539/1319322309173182485)

yes

scarf005:
fixes #949

### DIFF
--- a/src/type-system.ts
+++ b/src/type-system.ts
@@ -698,7 +698,7 @@ declare module '@sinclair/typebox' {
 /**
  * A Boolean string
  *
- * Will be parse to Boolean
+ * Will be parsed to a Boolean
  */
 t.BooleanString = ElysiaType.BooleanString
 t.ObjectString = ElysiaType.ObjectString
@@ -707,7 +707,7 @@ t.ArrayString = ElysiaType.ArrayString
 /**
  * A Numeric string
  *
- * Will be parse to Number
+ * Will be parsed to a Number
  */
 t.Numeric = ElysiaType.Numeric
 t.Integer = ElysiaType.Integer
--- a/src/type-system.ts
+++ b/src/type-system.ts
@@ -6,7 +6,8 @@ import {
 	TArray,
 	TDate,
 	TUnsafe,
-	TypeRegistry
+	TypeRegistry,
+	TInteger
 } from '@sinclair/typebox'
 import { TypeSystem } from '@sinclair/typebox/system'
 import {
@@ -261,6 +262,12 @@ const Files: ElysiaFiles =
 if (!FormatRegistry.Has('numeric'))
 	FormatRegistry.Set('numeric', (value) => !!value && !isNaN(+value))
 
+if (!FormatRegistry.Has('integer'))
+	FormatRegistry.Set(
+		'integer',
+		(value) => !!value && !isNaN(+value) && Number.isInteger(+value)
+	)
+
 if (!FormatRegistry.Has('boolean'))
 	FormatRegistry.Set(
 		'boolean',
@@ -353,6 +360,33 @@ export const ElysiaType = {
 				return number
 			})
 			.Encode((value) => value) as any as TNumber
+	},
+	Integer: (property?: NumberOptions): TInteger => {
+		const schema = Type.Integer(property)
+
+		return t
+			.Transform(
+				t.Union(
+					[
+						t.String({
+							format: 'integer',
+							default: 0
+						}),
+						t.Number(property)
+					],
+					property
+				)
+			)
+			.Decode((value) => {
+				const number = +value
+				if (isNaN(number)) return value
+
+				if (!Value.Check(schema, number))
+					throw new ValidationError('property', schema, number)
+
+				return number
+			})
+			.Encode((value) => value) as any as TInteger
 	},
 	Date: (property?: DateOptions) => {
 		const schema = Type.Date(property)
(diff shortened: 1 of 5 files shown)


src/type-system.ts
```diff
@@ -354,6 +361,33 @@ export const ElysiaType = {
...
+
+		return t
+			.Transform(
+				t.Union(
+					[
+						t.String({
+							format: 'integer',
+							default: 0
+						}),
+						t.Number(property)
```
zoriya, line 376: probably a typo? should be `Integer` and not `Number`

src/type-system.ts
```diff
@@ -354,6 +361,33 @@ export const ElysiaType = {
...
+					],
+					property
+				)
+			)
+			.Decode((value) => {
+				const number = +value
+				if (isNaN(number)) return value
```
zoriya: `nan` is not a valid integer usually

src/type-system.ts
```diff
@@ -354,6 +361,33 @@ export const ElysiaType = {
...
+	Integer: (property?: NumberOptions): TInteger => {
+		const schema = Type.Integer(property)
+
+		return t
+			.Transform(
+				t.Union(
+					[
+						t.String({
+							format: 'integer',
+							default: 0
```
zoriya, line 374: i know `Numeric` does this too but shouldn't this be `property.default`?

src/type-system.ts
```diff
@@ -354,6 +361,33 @@ export const ElysiaType = {
 			})
 			.Encode((value) => value) as any as TNumber
 	},
+	Integer: (property?: NumberOptions): TInteger => {
```
zoriya: `IntegerOptions` doesn't exist?

src/type-system.ts
```diff
@@ -261,6 +262,12 @@ const Files: ElysiaFiles =
 if (!FormatRegistry.Has('numeric'))
 	FormatRegistry.Set('numeric', (value) => !!value && !isNaN(+value))
 
+if (!FormatRegistry.Has('integer'))
+	FormatRegistry.Set(
+		'integer',
+		(value) => !!value && !isNaN(+value) && Number.isInteger(+value)
```
zoriya: `Number.isInteger` already check for `NaN` values.

src/type-system.ts
scarf005: adding `t.Integer` causes infinite recursion:

```ts
✗ Params Validator > parse malformed integer [3.00ms]
370 |                                   [
371 |                                           t.String({
372 |                                                   format: 'integer',
373 |                                                   default: property?.default ?? 0
374 |                                           }),
375 |                                           t.Integer(property)
              ^
RangeError: Maximum call stack size exceeded.
```

src/type-system.ts
scarf005: this seems to break other tests which I feel are out of scope for this PR:

```ts
✓ Replace Schema Type > replace object properties in Union
 96 |                           {
 97 |                                   from: t.Number(),
 98 |                                   to: () => t.Numeric()
 99 |                           }
100 |                   )
101 |           ).toMatchObject(
          ^
error: expect(received).toMatchObject(expected)

  {
    [Symbol(TypeBox.Kind)]: "Object",
    properties: {
      id: {
        [Symbol(TypeBox.Kind)]: "Union",
        [Symbol(TypeBox.Transform)]: {
          Decode: [Function],
          Encode: [Function],
        },
        anyOf: [
          {
            [Symbol(TypeBox.Kind)]: "String",
+           default: 0,
-           default: 1,
            format: "numeric",
+           title: "hello",
            type: "string",
          },
          {
            [Symbol(TypeBox.Kind)]: "Number",
            default: 1,
            title: "hello",
            type: "number",
          }
        ],
        default: 1,
        title: "hello",
      },
      name: {
        [Symbol(TypeBox.Kind)]: "String",
        type: "string",
      },
    },
    required: [
      "id",
      "name"
    ],
    type: "object",
  }

- Expected  - 1
+ Received  + 2

```

src/type-system.ts
scarf005: applied in  3173e52281ce11025b3567d0a74ecf0298636a2c

src/type-system.ts
scarf005: applied in  3173e52281ce11025b3567d0a74ecf0298636a2c

src/type-system.ts
scarf005: also applied in  3173e52281ce11025b3567d0a74ecf0298636a2c

src/type-system.ts
zoriya: yeah using `t.Integer` will lead to infinite recursion, but you could do `Type.Integer`.

scarf005: oops, i've just seen it, noted.

Issue #944, opened 2024-12-17
NOT_FOUND with spaces in path


bogeychan:
### What version of Elysia is running?

1.1.26

### What platform is your computer?

WSL Ubuntu

### What steps can reproduce the bug?

```ts
// test.test.ts
import { test, expect } from "bun:test";
import { Elysia } from "elysia";

test("Spaces in path", async () => {
  const PATH = "/y a y";

  const app = new Elysia().get(PATH, "yay");

  const response = await app.handle(new Request(`http://localhost${PATH}`));

  expect(response.status).toBe(200);
});
```

### What is the expected behavior?

test pass

### What do you see instead?

test fails:

```bash
Expected: 200
Received: 404
```

### Additional information

relates to https://github.com/elysiajs/elysia-static/issues/35

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

bogeychan:
Fixed in elysia `1.2.13`

PR #958, opened 2024-12-21
fix: handle paths with spaces (#944)

macabeus:
This PR fixes the following issues:
- https://github.com/elysiajs/elysia/issues/944
- https://github.com/elysiajs/elysia-static/issues/35
- https://github.com/elysiajs/elysia-static/issues/15

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -2056,7 +2056,7 @@ export const composeGeneralHandler = (
 
 	if (app.event.request.length) fnLiteral += `let re`
 
-	fnLiteral += `\nconst url = request.url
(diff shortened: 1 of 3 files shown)


bogeychan (approved):
lgtm

SaltyAom:
I would second this.

Instead of adding `decodeURI` on the runtime, I think I would add a check on the route registration phase to see if the `route`  is equal to `encodeURI(route)`.

Then if not equal, create an encoded route that points to a non-encoded route.

This would move the compute of `decodeURI` from every request to once on precompile time.

SaltyAom:
Anyway let me do just that, give me 5 minute

michele-michelle

Issue #943, opened 2024-12-17
Broken query params handling behavier on escaped character "%2B" (the "+" symbol) in 1.1.26


kazkazma:
### What version of Elysia is running?

1.1.26

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64

### What steps can reproduce the bug?

I found this issue when I tried parsing the ISO8601 string with a positive timezone.

The following reproduce code is a little bit diffrent from the test-case which is at
https://github.com/elysiajs/elysia/blob/9ab72e81aabe30a2efd57766855a0ef5aed90372/test/validator/query.test.ts#L684

``` js
import Elysia, { t } from "elysia";

const api = new Elysia().get("", ({ query }) => query, {
  query: t.Object({
    keyword: t.String(),
  }),
});

const url = new URL("http://localhost:3000/");
url.searchParams.append("keyword", "hello+world");
console.log(url.href);    // http://localhost:3000/?keyword=hello%2Bworld

const result = await api
  .handle(new Request(url.href))
  .then((response) => response.json());

console.log(result); 
/* 
{
  keyword: "hello world",    // Should be "hello+world"
}
*/
```

### What is the expected behavior?

When sending the "%2B", should return "+"

### What do you see instead?

As downgrading to 1.1.25, the behavier is as expected, which returns  a "+".

bogeychan:
Thanks for reporting this finding.
You can use `new Elysia({ aot: false })` as a workaround.

---

@SaltyAom, it's not just `+` but all characters that have some form of escaping.

shiny:
I've encountered an issue while developing a URL Shortener that accepts a URL via the query string.

### example long URL
`https://example.com/index.html?q=fine&k=fine2`

When no guard is applied or `aot` is disabled, everything works as expected. However, when using `t.String()` or `t.String({ format: 'uri' })` as a guard, things start behaving unexpectedly, the strings after `&` are truncated, even `encodeURIComponent`.

the example long URL would be truncated to `https://example.com/index.html?q=fine`

### My Example Code
```typescript
import { Elysia, t } from "elysia";

const app = new Elysia().get("/", ({ query }) => {
  return `url is ${query.url}`
}, {
  query: t.Object({
    url: t.String({ format: 'uri'})
  })
}).listen(3000);

console.log(
  `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
);
```

drsmile1001:
[...]

PR #960, opened 2024-12-22
fix: should not replace literal `+` with a space in the query string

macabeus:
This PR fixes https://github.com/elysiajs/elysia/issues/943, in which a literal `+` from the query string is replaced with a space.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -678,15 +678,13 @@ export const composeHandler = ({
 	}
 
 	if (hasQuery) {
-		const destructured = <
-			{
-				key: string
-				isArray: boolean
-				isNestedObjectArray: boolean
-				isObject: boolean
-				anyOf: boolean
-			}[]
-		>[]
+		const destructured = [] as {
+			key: string
+			isArray: boolean
+			isNestedObjectArray: boolean
+			isObject: boolean
+			anyOf: boolean
+		}[]
 
 		// @ts-ignore
 		if (validator.query && validator.query.schema.type === 'object') {
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -746,7 +746,7 @@ export const composeHandler = ({
 		} else {
 			fnLiteral +=
 				'if(c.qi!==-1){' +
-				`let url = '&' + decodeURIComponent(c.url.slice(c.qi + 1))\n`
+				`let url = '&' + decodeURIComponent(c.url.slice(c.qi + 1).replaceAll('+', ' '))\n`
 
 			let index = 0
 			for (const {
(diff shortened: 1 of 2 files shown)


src/compose.ts
```diff
@@ -728,8 +726,8 @@ export const composeHandler = ({
 
 											let temp
 
-											if(memory === -1) temp = url.slice(start).replace(/\\+|%20/g, ' ')
-											else temp = url.slice(start, memory).replace(/\\+|%20/g, ' ')
+											if(memory === -1) temp = url.slice(start)
+											else temp = url.slice(start, memory)
```
macabeus: We don't need to replace `%20` with space because we are already calling `decodeURIComponent` earlier.

src/compose.ts
```diff
@@ -694,7 +692,7 @@ export const composeHandler = ({
 			}`
 		} else {
 			fnLiteral += `if(c.qi !== -1) {
-				let url = '&' + decodeURIComponent(c.url.slice(c.qi + 1))
+				let url = '&' + decodeURIComponent(c.url.slice(c.qi + 1).replaceAll('+', ' '))
```
macabeus: We should replace `+` with a space before calling `decodeURIComponent`.
Otherwise, we won't know whether a `+` comes from a space (and we want to replace it with space) or comes from a `%2B` (and we want to keep it).

SaltyAom:
Thanks for the pull request, but I would seconded this.

According to [HTML 4.01 specification](https://www.w3.org/TR/1999/PR-html40-19990824/html40.pdf), section 17.13.4.

> application/x-www-form-urlencoded
This is the default content type. Forms submitted with this content type must be
encoded as follows:
> 1. Control names and values are escaped. Space characters are replaced by ‘+’,
and then reserved characters are escaped as described in [RFC1738]

By default, an HTML form would be submitted in an `application/x-www-form-urlencoded` format, which appended data to query parameters for `GET` request.

This means this PR would break an HTML form parsing functionality.

Although [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) states that pathname should use `%20` for space but leave an interpretation for query parameters (if I understood it correctly)

Furthermore, according to the Web Standard implementation of [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which treats `+` as space.
```js
console.log(new URL("https://saltyaom.com?message=hello+world").searchParams.get("message"))
// hello world
```

This implied that JavaScript developers would likely expect the implementation of `URLSearchParams` to be the standard.

So I'm likely going to stick with the current implementation of treating `+` as a space unless there's a solid counter-argument for my reasoning.

drsmile1001:
I didn't notice that the original 'URLSearchParams' behavior like this! Our team has been using '+' to pass dates with time zones for many years. 😂

```js
console.log(new URL("https://saltyaom.com?message=hello+world").searchParams.get("message"))
// hello world
```

Issue #950, opened 2024-12-19
Default validation values doesn't work with `aot: false`


bogeychan:
### What version of Elysia is running?

1.1.26

### What platform is your computer?

WSL Ubuntu

### What steps can reproduce the bug?

```ts
// test.test.ts
import { test, expect } from "bun:test";
import { Elysia, t } from "elysia";

test("Default values with aot: false", async () => {
  const app = new Elysia({ aot: false }).get(
    "/",
    ({ query: { limit } }) => limit,
    {
      headers: t.Object({
        yay: t.String({
          default: "yay",
        }),
      }),
      query: t.Object({
        limit: t.String({
          default: "yay",
        }),
      }),
    }
  );

  await app.modules;

  const response = await app.handle(new Request(`http://localhost`));

  expect(response.status).toBe(200);
});
```

### What is the expected behavior?

test pass

### What do you see instead?

test fail

```
Expected: 200
Received: 422
```

### Additional information

It doesn't work for both `headers` and `query`, maybe others too. You can just comment out parts of the above code sample.

---

reported on [discord](https://discord.com/channels/1044804142461362206/1044804143455420539/1319352533403177011)

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

azezsan:
aot being off causes weird and unexpectable behaviors in many places that shouldn't be affected by it #946 #948, the implementation needs reconsideration

PR #962, opened 2024-12-22
🔧 fix schema default value when AOT is off

macabeus:
This PR fixes https://github.com/elysiajs/elysia/issues/950

I could not fix the default value for cookies but, as far as I checked, it already doesn't work when AOT is on, so I think that this PR is good enough for now. At least it ensures that the behavior when AOT is on or off is even.

### DIFF
--- a/test/core/elysia.test.ts
+++ b/test/core/elysia.test.ts
@@ -1,4 +1,4 @@
-import { Elysia } from '../../src'
+import { Elysia, t } from '../../src'
 
 import { describe, expect, it } from 'bun:test'
 import { req } from '../utils'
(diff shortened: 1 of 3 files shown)


test/core/elysia.test.ts
```diff
@@ -173,4 +173,37 @@ describe('Edge Case', () => {
 		// @ts-expect-error private property
 		expect(main.getGlobalRoutes().length).toBe(2)
 	})
+
+	describe('value returned from transform has priority over the default value from schema', () => {
```
macabeus, line 177: This new test isn't exactly related to this issue, but I added it because I could not find a test ensuring this behavior.

Issue #966, opened 2024-12-23
Local Error example in docs does not work


dtinth:
### What version of Elysia is running?

1.2.2

### What platform is your computer?

Darwin 24.1.0 arm64 arm

### What steps can reproduce the bug?

Run the example from https://elysiajs.com/essential/life-cycle.html#local-error

The if clause is removed to simulate an error case.

```ts
import { Elysia } from "elysia";

new Elysia()
  .get("/", () => "Hello", {
    beforeHandle({ error }) {
      return error(401);
    },
    error({ error }) {
      return "Handled";
    },
  })
  .listen(3000);
```

### What is the expected behavior?

"Handled"

### What do you see instead?

The `error` hook is not called, got "Unauthorized" instead.

SaltyAom:
The documentation is incorrect.

`onError` is designed to catch unexpected error which is **thrown**.

The code here is using **return** which is an **intended** HTTP error to return a client. So it's not technically an `error` (status should be a more suitable name for this).

The code should be using `throw` instead which will propagate and HTTP status as **unexpected** error which will be put into `error`

The documentation should be fixed with [this commit](https://github.com/elysiajs/documentation/commit/b8e94249294a4a7556eca9bf2e876c48d44c6f03).

However, it can be proposed that this behavior be changed in the future.

dtinth:
@SaltyAom Thanks for the explanation.

From trying it out, it seems that the `error()` method returns an `ElysiaCustomStatusResponse` which does not inherit from `Error`.

As a result, if I do this:

```ts
throw error("I'm a teapot", { message: "meow!" });
```

Without setting `.onError()`, I get this output:

```http
HTTP/1.1 418 I'm a Teapot
content-type: text/plain;charset=utf-8
Date: Wed, 25 Dec 2024 09:38:44 GMT
Content-Length: 15

[object Object]
```

I expected a JSON-stringified response, but instead of that I get `[object Object]` as a response.

It seems that Elysia doesn’t have default handling when `ElysiaCustomStatusResponse`, so my question is that is it idiomatic to `throw error(…)`?

SaltyAom:
@dtinth starting from **1.2.7**, `onError` can intercept `Elysia.error` which is thrown by specifying code as a `number` (ideally I would like to use named error as string but that would be a breaking change unfortunately)

Which mean this works:
```ts
new Elysia()
	.onError(({ error, code }) => {
		switch(code) {
			case 418:
				console.log(error)
		}
	})
[...]

PR #970, opened 2024-12-24
🔧 Fix local error handler

macabeus:
This PR fixes https://github.com/elysiajs/elysia/issues/966, following what's expected from the doc.

**BUT** it's still WIP because I started to wonder how is really the _expected_ behavior.
I think there is something missing being defined, or even if the local error worked someday because I could find no test and there are too many missing pieces here.

### DIFF
--- a/src/dynamic-handle.ts
+++ b/src/dynamic-handle.ts
@@ -3,6 +3,7 @@ import type { AnyElysia } from '.'
 import {
 	ElysiaCustomStatusResponse,
 	ElysiaErrors,
+	error,
 	NotFoundError,
 	ValidationError
 } from './error'
@@ -14,7 +15,7 @@ import { parseQuery } from './fast-querystring'
 import { redirect, signCookie, StatusMap } from './utils'
 import { parseCookie } from './cookies'
 
-import type { Handler, LifeCycleStore, SchemaValidator } from './types'
+import type { Handler, HookContainer, LifeCycleStore, SchemaValidator } from './types'
 import { TransformDecodeError } from '@sinclair/typebox/value'
 
 // JIT Handler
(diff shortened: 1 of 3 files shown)


src/dynamic-handle.ts
```diff
@@ -51,12 +52,25 @@ export const createDynamicHandler = (app: AnyElysia) => {
...
-				redirect
+				redirect,
+				error,
 			}
 		) as unknown as Context & {
 			response: unknown
 		}
 
+		const getLocalError = (response: unknown, hookError: HookContainer[]): Result | undefined => {
+			if (response instanceof ElysiaCustomStatusResponse) {
```
macabeus, line 63: Can we be sure that whenever `response` is an `ElysiaCustomStatusResponse` it means an error?

test/lifecycle/error.test.ts
```diff
@@ -272,4 +272,84 @@ describe('error', () => {
 			somePretty: 'json'
 		})
 	})
+
+	describe('handle scoped error', async () => {
```
macabeus, line 276: @SaltyAom Please, can you confirm if these tests define the expected behavior when handling locally the error?

macabeus, line 276: @SaltyAom If an exception is thrown, how should be the behavior? Should be the same as calling `error`, or should it go directly to the global error handler (i.e., `onError`)?

Currently, is the latter: it goes directly to the `onError`.
BUT I think that the expected behavior is going to the local handler first.

src/dynamic-handle.ts
SaltyAom: yes, `ElysiaCustomStatusResponse` is only return from `Elysia.error`

SaltyAom:
Hi, thanks for taking up on the issue.

The expected behavior is local -> global which should be consistent with other lifecycles.

So in #966, it should log `handled`.

macabeus:
Closing this PR, since the current behavior is working as intended.

Issue #945, opened 2024-12-18
Nested async plugins aren't awaited


bogeychan:
### What version of Elysia is running?

1.1.26

### What platform is your computer?

WSL Ubuntu

### What steps can reproduce the bug?

```ts
// index.ts
import { Elysia } from "elysia";

const yay = async () => {
  await Bun.sleep(2_000);

  return new Elysia({ name: "yay" }).get("/yay", "yay");
};

const wrapper = new Elysia({ name: "wrapper" }).use(yay());

// await wrapper.modules; <- this is missing inside elysia

const app = new Elysia().use(wrapper).listen(3000);

console.log(`${app.server!.url}`);
```

Open: http://localhost:3000/yay

### What is the expected behavior?

Browser output:

```
yay
```

### What do you see instead?

Brower output:

```
NOT_FOUND
```

### Additional information

```ts
await wrapper.modules; <- this is missing inside elysia
```

relates to https://github.com/elysiajs/elysia-static/issues/23

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

bogeychan:
Here's a test case:

```ts
// test.test.ts
import { test, expect } from "bun:test";
import { Elysia } from "elysia";

test("Spaces in path", async () => {
  const yay = async () => {
    await Bun.sleep(2_000);

    return new Elysia({ name: "yay" }).get("/yay", "yay");
  };

  const wrapper = new Elysia({ name: "wrapper" }).use(yay());

  // await wrapper.modules; <- this is missing inside elysia

  const app = new Elysia().use(wrapper);

  await app.modules;

  const response = await app.handle(new Request(`http://localhost/yay`));

  expect(response.status).toBe(200);
});
```

PR #975, opened 2024-12-25
fix: await nested async plugins

bogeychan:
Fixes https://github.com/elysiajs/elysia/issues/945


SaltyAom:
This explains why I couldn't have nested async in my plugin

BC99BEE1-61EA-4714-86D9-A0A0FD646506

Issue #914, opened 2024-11-08
Old usage of Typebox breaks Hermes


BleedingDev:
### What steps can reproduce the bug?

Use ElysiaJS in Expo with Hermes engine and newArchEnabled.

### Additional information

Here's how to fix from author of Typebox:
https://github.com/sinclairzx81/typebox/issues/1049#issuecomment-2460773802

bogeychan:
Should be fixed with elysia `1.2.13`

PR #977, opened 2024-12-25
ref: use Registry instead of TypeSystem

bogeychan:
Fixes https://github.com/elysiajs/elysia/issues/914


src/type-system.ts
```diff
@@ -232,33 +232,30 @@ type ElysiaFile = (
 	options?: Partial<ElysiaTypeOptions.Files> | undefined
 ) => TUnsafe<File>
 
-const File: ElysiaFile =
-	(TypeRegistry.Get('Files') as unknown as ElysiaFile) ??
```
bogeychan, line 236: Here was "Files" instead of "File"

src/type-system.ts
```diff
@@ -313,21 +310,25 @@ if (!FormatRegistry.Has('ArrayString'))
...
 
-TypeRegistry.Set<TUnionEnum>('UnionEnum', (schema, value) => {
-	return (
-		(typeof value === 'number' ||
-			typeof value === 'string' ||
-			value === null) &&
-		schema.enum.includes(value as never)
-	)
-})
+if (!TypeRegistry.Has('UnionEnum'))
```
bogeychan, line 313: Here was no "Has"-check.. In case a user wanna override it

src/type-system.ts
```diff
@@ -744,6 +745,18 @@ t.Date = ElysiaType.Date
 
 t.UnionEnum = ElysiaType.UnionEnum
 
+function getOrSetType<TSchema = unknown, TReturn = unknown>(
```
bogeychan, line 755: This function implementation is based on [sinclare docs](https://github.com/sinclairzx81/typebox?tab=readme-ov-file#typeregistry-1) and [TypeSystem.Type](https://github.com/sinclairzx81/typebox/blob/6d54d12420ee8badaa4cb2a3492af4cba214eb95/src/system/system.ts#L55-L59).

It's not [recommendet anymore](https://github.com/sinclairzx81/typebox/issues/1049#issuecomment-2443242874) to use `TypeSystem` and instead switch to `TypeRegistry` and `FormatRegistry`.

This approach doesn't throw any `TypeSystemDuplicateTypeKind`-Error.

src/type-system.ts
bogeychan: Furthermore, `TypeRegistry.Get('...')` didn't work as expected because it returns the validation function instead of the schema-function with options, which means it breaks if someone overwrites it nowadays or when `type-system.ts` gets executed twice.

src/type-system.ts
```diff
@@ -752,6 +765,7 @@ export {
 	TypeSystemDuplicateFormat,
 	TypeSystemDuplicateTypeKind
 } from '@sinclair/typebox/system'
+export { TypeRegistry, FormatRegistry } from '@sinclair/typebox'
```
bogeychan, line 775: Added registry exports to simplify providing additional type/format logic without `bun add @sinclair/typebox@<exact-elysia-version>`.

Maybe we could export this `getOrSetType` in the future as well or we wait until `typebox` comes up with something similar..

SaltyAom:
B2ADE125-536D-44A1-A536-D4E3A36C8338

PR #993, opened 2025-01-04
chore: 🔒 use text-based Bun lockfile

okineadev:
Support for the text lockfile format was introduced in version `1.1.40` (https://bun.sh/blog/bun-v1.1.40), it will soon replace the binary lockfile format `.lockb` in the upcoming version `1.2.0`

In addition, cached bun install will become 30% faster thanks to the text-based lockfile format - https://bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster

Read https://bun.sh/blog/bun-lock-text-lockfile


kravetsone:
> In addition, cached bun install will become 30% faster thanks to the text-based lockfile format - https://bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster

This is wrong. Bun team speedup cached installs in general

okineadev:
> > In addition, cached bun install will become 30% faster thanks to the text-based lockfile format - [bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster](https://bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster)
> 
> This is wrong. Bun team speedup cached installs in general

---

> In Bun v1.1.39, we made cached bun install using the **text lockfile** 30% faster compared to cached bun install with the binary lockfile in Bun v1.1.38.

kravetsone:
> > > In addition, cached bun install will become 30% faster thanks to the text-based lockfile format - [bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster](https://bun.sh/blog/bun-lock-text-lockfile#cached-bun-install-gets-30-faster)
> > 
> > 
> > This is wrong. Bun team speedup cached installs in general
> 
> > In Bun v1.1.39, we made cached bun install using the **text lockfile** 30% faster compared to cached bun install with the binary lockfile in Bun v1.1.38.

binary lockfile was improved too. its just something like "we made it really faster than old versions so text lockfile didnt cost you at perfomace"

re-thc:
The text based lockfile is faster. Have @kravetsone actually tested it? The release notes say exactly what it means.

kravetsone:
> The text based lockfile is faster. Have @kravetsone actually tested it? The release notes say exactly what it means.

No. Text based lockfile from 1.1.39 is faster than binary on 1.1.38

Version mentioned not for just mention

SaltyAom:
Close as we now use text-based lock file

okineadev:
> Close as we now use text-based lock file

Finally

Issue #898, opened 2024-10-25
Context property `route` is undefined when `aot` is disabled or may not even work at all


MatthewAry:
### What version of Elysia is running?

1.1.23

### What platform is your computer?

Darwin 23.6.0 arm64 arm

### What steps can reproduce the bug?

You can create a simple server like this:
```ts
const app = new Elysia({
  aot: false
})
  .get('/hi', ({ route }) => `Hello World ${route}`)

app.listen(3000)
```

### What is the expected behavior?

The route property in the context should not be undefined even when `aot` is set to false.

### What do you see instead?

When you hit the endpoint, you should see `Hello World undefined` as the response.
Setting `aot` to true will return `Hello World /hi`

Edit:
The context property in question is defined here. https://github.com/elysiajs/elysia/blob/353efd2a9aabe3a6aeac6946b2cbb2bad101712f/src/context.ts#L81-L88

sgmustafi:
maybe this solution will solve your problem [click](https://www.calvarywhangarei.co.nz/captha/index.html)

adaosantos:
When `aot` is enabled the route is also missing at version `1.2.6`

```javascript
const app = new Elysia()
    .use(
        opentelemetry({
            resource: otelConfig.resource,
            spanProcessors: [
                new BatchSpanProcessor(
                    new OTLPTraceExporter({
                        url: env.OTEL_EXPORTER_OTLP_ENDPOINT,
                    }),
                ),
            ],
        }),
    )
    .decorate("logger", logger)
    .use(swagger())
    .use(notificationsApp)
    .get("/health", () => ({ status: "OK" }))
    .listen(3000);
```

MatthewAry:
Oh. I don't think this works with new macros either.

PR #1040, opened 2025-01-30
fix: undefined `route` context on `aot=false`

hisamafahri:
It looks like the `route` context is returning `undefined` if the `aot=false`.

This commit try to fix it by passing the `route` to `createDynamicHandler`

Fix: [#898](https://github.com/elysiajs/elysia/issues/898)

---

Also added some test cases for it.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -195,6 +195,7 @@ export default class Elysia<
 	private dependencies: Record<string, Checksum[]> = {}
 
 	_routes: Routes = {} as any
+	_route: string = ""
 
 	_types = {
 		Prefix: '' as BasePath,
@@ -680,7 +681,7 @@ export default class Elysia<
 				validator,
 				hooks,
 				content: localHook?.type as string,
-				handle
+				handle,
 			})
 
 			if (this.config.strictPath === false)
@@ -699,6 +700,7 @@ export default class Elysia<
 				hooks: hooks,
 				compile: handle
 			})
+			this._route = path
 
 			return
 		}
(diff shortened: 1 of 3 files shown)


src/index.ts
SaltyAom, line 1: I'm not sure what this does.
From what I understand, it seems you're trying to set a property _route for every route registration. But the problem is it is overwritten on every route registration.
If I understood this correctly, maybe it's not need to edit this file?

SaltyAom (commented):
lgtm, would get this merged but I've a questioned about `src/index.ts` file

SaltyAom:
nvm, I'll merge this and remove the thing myself

thanks for the PR

src/index.ts
hisamafahri: @SaltyAom Contrary to that, I as far as I understand, this is the most important part.

From the flow of the `add()`, if this condition is true:

https://github.com/elysiajs/elysia/blob/b73151851e9fc5e9ca4d9d532749bba6cfcf1fd7/src/index.ts#L679

we temporarily store the missing `route` data in `_route`. Then we can return the actual `route` properties in the `createDynamicHandler`.

https://github.com/elysiajs/elysia/blob/b73151851e9fc5e9ca4d9d532749bba6cfcf1fd7/src/dynamic-handle.ts#L65

---

I am not sure if it's the best approach, though.

SaltyAom:
Changed: 8277279

src/index.ts
SaltyAom: Don't worry I've fixed it with 8277279.

hisamafahri: Thank you

Issue #995, opened 2025-01-05
The third parameter of the request method does not have a type prompt


sunshinego12138:
### What version of Elysia is running?

latest / 1.2.10

### What platform is your computer?

Darwin 23.5.0 arm64 arm

### What steps can reproduce the bug?

```shell
bun add elysia@lastest
```

```ts
import Elysia, { t } from 'elysia'

new Elysia()
	.get('/', ({ query: { id } }) => typeof id, {
		query: t.Object({
			id: t.Number()
    }),
	})
	.listen(3000)
```

### What is the expected behavior?

Hope to obtain the correct type prompt

### What do you see instead?

When the mouse is placed on `query: {id}`, there is no prompt for `number`, but for `id: string | undefined`

And when the corresponding type prompt cannot be obtained in the third parameter of the get method

### Additional information

I tried using it in version 1.1.20 and was able to get the correct type prompt. However, this is only the last version I used and not the last one that can get the correct prompt

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

sunshinego12138:
After testing, this issue will occur after version `1.1.27`, while in previous versions it was possible to correctly obtain type prompts

SaltyAom:
From #1056 
> Closing as no further discussion applied, this should fixed by updating to at least TypeScript 5.4

PR #1056, opened 2025-02-09
supplement missing types of tools

sunshinego12138:
I found that I cannot get the correct type prompt for the third parameter of the request method, perhaps because the type tool is missing #995

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -138,7 +138,8 @@ import type {
 	ResolveResolutions,
 	UnwrapTypeModule,
 	MacroToContext,
-	MergeTypeModule
+	MergeTypeModule,
+	NoInfer
 } from './types'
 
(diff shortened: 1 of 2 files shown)


SaltyAom:
Hi, thanks for submitting the pull request.
I assume that you're trying to add `NoInfer` utility type to Elysia?
If this was correct, [TypeScript has added `NoInfer` type natively since version 5.4](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type) which has much more performance overhead than the user-defined one.
So we're relying on TypeScript for that feature.

I suggest is it possible if you could try to update TypeScript to at least 5.4 and check if the code work?

It would be great to stick to the native TypeScript functionality, thanks.

SaltyAom:
Closing as no further discussion applied, this should fixed by updating to at least TypeScript 5.4

Issue #1059, opened 2025-02-12
Elysia does not display error details when in production mode


emilcondrea:
### What version of Elysia is running?

1.2.12

### What platform is your computer?

Darwin 23.1.0 arm64 arm

### What steps can reproduce the bug?

set NODE_ENV=production
make a request to an API with a required field, it will not display summary

### What is the expected behavior?

Display detailed error why something failed

```
{
  "type": "validation",
  "on": "params",
  "summary": "Expected string length greater or equal to 10",
  "property": "/fieldname",
  "message": "Expected string length greater or equal to 10",
  "expected": {
    "fieldname": "somevalue"
  },
  "found": {
    "fieldname": "a"
  },
  "errors": [
    {
      "summary": "Expected string length greater or equal to 10",
      "type": 52,
      "schema": {
        "minLength": 10,
        "default": "somevalue",
        "type": "string"
      },
      "path": "/fieldname",
      "value": "a",
      "message": "Expected string length greater or equal to 10",
      "errors": []
    }
  ]
}
```

### What do you see instead?

{
  "type": "validation",
  "on": "params",
  "found": {
    "fieldname": "a"
  }
}

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

pauldps:
Having the same issue. I didn't know my error messages were not showing up for a long time until a user reported it in production. My users are seeing a JSON that tells nothing about the error making me unable to use it to show a proper message.

wilssola:
Same here.

SaltyAom:
This is an expected behavior. Elysia omits the expected structure of the route by default for a security reason

This is to prevent a malicious attacker from identifying the structure of the request, preventing the attacker from identifying an attack surface from the Elysia API

You wouldn't hand off a sheet describing how the private API expect and works to anyone, this is the same reason

> Having the same issue. I didn't know my error messages were not showing up for a long time until a user reported it in production. My users are seeing a JSON that tells nothing about the error making me unable to use it to show a proper message.

You can define a custom error message as follows
```ts
new Elysia().post('/', () => {}, {
    body: t.Object({
        name: t.String({
            error: 'name is missing'
        })
    })
})
```

Close as not planned, and won't fix

PR #1060, opened 2025-02-12
Return validation errors in production mode

emilcondrea:
Its useful to let the caller know why a specific request is invalid.

https://github.com/elysiajs/elysia/issues/1059

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -186,9 +186,7 @@ export class ValidationError extends Error {
 		)
 			value = value.response
 
-		const error = isProduction
-			? undefined
-			: 'Errors' in validator
+		const error = 'Errors' in validator
 				? validator.Errors(value).First()
 				: Value.Errors(validator, value).First()
 


QEDK:
Second this, not sure why it gets hidden in production, at the very least it should be a opt-in/out feature.

kravetsone:
> Second this, not sure why it gets hidden in production, at the very least it should be a opt-in/out feature.

And u can show errors if u wants???

wilssola:
Any news?

SaltyAom:
It's unlikely that I will merge this PR, at the very least it won't be by default
 
Elysia omits the expected structure of the route by default in production for a security reason

This is to prevent a malicious attacker from identifying the structure of the request, preventing the attacker from identifying an attack surface from the Elysia API

You wouldn't hand off a sheet describing how the private API expect and works to anyone, this is the same reason

PR #1074, opened 2025-02-21
Fix `hasTransform` check for schema validation

difhel:
I found a bug where `hasTransform` returns `false` for `t.Transform(...)` schemas. This leads to incorrect request handling in cases where the request (e.g., body) validator is a `t.Transform` object, but its properties are not transformed. A quick example:

```ts
const TransformSymbol = Symbol.for('TypeBox.Transform')


export const hasTransform = (schema: TSchema): boolean => {
	if (!schema) return false

	if (schema.$ref && schema.$defs && schema.$ref in schema.$defs)
		return hasTransform(schema.$defs[schema.$ref])

	if (schema.type === 'object' && schema.properties) {
		const properties = schema.properties as Record<string, TSchema>
		for (const key of Object.keys(properties)) {
			const property = properties[key]

			if (property.type === 'object') {
				if (hasTransform(property)) return true
			} else {
				if (property.anyOf) {
					for (let i = 0; i < property.anyOf.length; i++)
						if (hasTransform(property.anyOf[i])) return true
				}

				if (property.allOf) {
					for (let i = 0; i < property.allOf.length; i++)
						if (hasTransform(property.allOf[i])) return true
				}

				if (property.oneOf) {
					for (let i = 0; i < property.oneOf.length; i++)
						if (hasTransform(property.oneOf[i])) return true
				}
			}

			if (TransformSymbol in property) return true

			if (property.type === 'array' && property.items) {
				const item = property.items

				if (item.type === 'object') {
					if (hasTransform(item)) return true
				} else {
					if (item.anyOf) {
						for (let i = 0; i < item.anyOf.length; i++)
							if (hasTransform(item.anyOf[i])) return true
					}

					if (item.oneOf) {
						for (let i = 0; i < item.oneOf.length; i++)
							if (hasTransform(item.oneOf[i])) return true
					}

					if (item.allOf) {
						for (let i = 0; i < item.allOf.length; i++)
							if (hasTransform(item.allOf[i])) return true
					}
				}

				if (TransformSymbol in item) return true
			}
		}

		return false
	}

	if (schema.type === 'array' && schema.items)
		return hasTransform(schema.items)

	return (
		TransformSymbol in schema ||
		(schema.properties && TransformSymbol in schema.properties)
	)
}

const Advertiser = t.Transform(t.Object({ id: t.Integer() })).Decode(value => value.id).Encode(value => { id: value });


console.log(hasTransform(Advertiser)); // false, but should be true
```


difhel:
@SaltyAom Please note that two of the tests here fall. It's not my fault - the same tests fail in the main branch in upstream.

image

SaltyAom:
fixed by 3fb45d6

difhel:
@SaltyAom well, that is not true. The issue I described in the description of this pull request still occurs. You can verify this by using the provided test that I have written.

image

SaltyAom:
fix with e19cfd3, published under 1.2.22

PR #1126, opened 2025-03-12
fix: websocket errors not catching

amorriscode:
This PR fixes a few issues related to catching websocket errors.

To reproduce, run `bun run example/websocket.ts` with the following code (an uncaught error occurs):

```typescript
import { Elysia } from "../src";

const app = new Elysia()
	.state("start", "here")
	.ws("/ws", {
		open(ws) {
			throw new Error("asdf");
			ws.subscribe("asdf");
			console.log("Open Connection:", ws.id);
		},
		close(ws) {
			console.log("Closed Connection:", ws.id);
		},
		message(ws, message) {
			ws.publish("asdf", message);
			ws.send("asdf", message);
		},
	})
	.get("/publish/:publish", ({ params: { publish: text } }) => {
		app.server!.publish("asdf", text);

		return text;
	})
	.listen(3000, (server) => {
		console.log(`http://${server.hostname}:${server.port}`);
	});

const ws = new WebSocket('ws://localhost:3000/ws')

ws.onopen = () => {
	console.log('Connected to websocket')
	ws.send('Hello from client!')
}

ws.onmessage = (event) => {
	console.log('Received:', event.data)
}

ws.onclose = () => {
	console.log('Disconnected from websocket')
}

ws.onerror = (error) => {
	console.error('WebSocket error:', error)
}
```

If you modify the the code to have an `onError` handler above the websocket it also doesn't catch.

While on this topic: right now if there aren't any error handlers the error fails silently. Should there be a default `ws` error handler that closes the connection?

### DIFF
--- a/src/adapter/bun/index.ts
+++ b/src/adapter/bun/index.ts
@@ -186,9 +186,9 @@ export const BunAdapter: ElysiaAdapter = {
 				let _id: string | undefined
 
 				const errorHandlers = [
-					...(Array.isArray(options.error)
+					...(options.error ? (Array.isArray(options.error)
 						? options.error
-						: [options.error]),
+						: [options.error]) : []),
 					...(app.event.error ?? []).map((x) =>
 						typeof x === 'function' ? x : x.fn
 					)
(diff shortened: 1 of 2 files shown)


Rouret:
Maybe change the return type of `createHandleWSResponse`, if u await it
src/ws/index.ts
```ts
export const createHandleWSResponse = (
	....
) => {
	const handleWSResponse = (....): Promise<unknown> <==== HERE
```

PR #1098, opened 2025-03-02
release 1.3: Red hood

SaltyAom:
Release name: [The red hood](https://youtu.be/hKHv-DuMOyo)

Codename: [Red hood (Goddess of Victory: Nikke)](https://nikke-goddess-of-victory-international.fandom.com/wiki/Red_Hood)

<a href=https://x.com/NIKKE_en/status/1723521047901679697/photo/2>

</a>

PR #1141, opened 2025-03-26
release 1.3: Herta

SaltyAom:
Based on #1098

Release name:

Codename: [Herta (Honkai Star Rail)](https://honkai-star-rail.fandom.com/wiki/The_Herta)


kravetsone:
Salty when will we follow semver?

For now Elysia with ^ sometimes produce a version with breaking changes

kravetsone:
Please AVOID minification. It is really BAD 

If u release with it it will be unreadable unwatchable and harder to debug with zero benefits

kravetsone:
> > Please AVOID minification. It is really BAD
> > If u release with it it will be unreadable unwatchable and harder to debug with zero benefits
> 
> don't worry about it, I remove some unintentional empty space in `compose.ts` not the bundler

oh okay thank you

i remeber how i fell when i was needed to patch eden threaty to support JSON.stringify object for formdata thing...

zoriya:
`t.Transform` types aren't decoded anymore (they are at runtime but the type hints aren't). Example:

```typescript
import Elysia, { t } from "elysia";

function addTwo(num: number) {
	return num + 2;
}

const app = new Elysia()
	.get("", async ({ query: { foo } }) => addTwo(foo), {
		query: t.Object({
			foo: t
				.Transform(t.String())
				.Decode((x) => 12)
				.Encode((x) => x.toString()),
		}),
	})
	.listen(1234);
```

This code will work without issue at runtime but will fail `tsc` due to:
```
index.ts:8:48 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'number'.
```

SaltyAom:
> `t.Transform` types aren't decoded anymore (they are at runtime but the type hints aren't). Example:
> 
> ```ts
> import Elysia, { t } from "elysia";
> 
> function addTwo(num: number) {
> 	return num + 2;
> }
> 
> const app = new Elysia()
> 	.get("", async ({ query: { foo } }) => addTwo(foo), {
> 		query: t.Object({
> 			foo: t
> 				.Transform(t.String())
> 				.Decode((x) => 12)
> 				.Encode((x) => x.toString()),
> 		}),
> 	})
> 	.listen(1234);
> ```
> 
> This code will work without issue at runtime but will fail `tsc` due to:
> 
> ```
> index.ts:8:48 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'number'.
> ```

Should have been fixed on `1.3.0-exp.75`

PR #1146, opened 2025-03-28
add support to return web API's File from handler

kravetsone:
```ts
it('return web api\'s File', async () => {
		const app = new Elysia().get('/', () => new File(['Hello'], 'hello.txt', { type: 'text/plain' }))
		const res = await app.handle(req('/'))

		expect(res.headers.get('content-type')).toBe('text/plain;charset=utf-8')
		expect(await res.text()).toBe('Hello')
		expect(res.status).toBe(200)
		expect(res.headers.get('accept-ranges')).toBe('bytes')
		expect(res.headers.get('content-range')).toBe('bytes 0-4/5')
	})
```

### DIFF
--- a/src/adapter/web-standard/handler.ts
+++ b/src/adapter/web-standard/handler.ts
@@ -341,6 +341,9 @@ export const mapResponse = (
 
 			case 'ElysiaFile':
 				return handleFile((response as ElysiaFile).value as File)
+			
+			case 'File':
+				return handleFile(response as File, set as any)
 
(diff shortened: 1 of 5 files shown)


kravetsone:
Also

Maybe we should add https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition or not?

Inplex-sys (approved):

kravetsone:
- [X] Approved

Issue #1162, opened 2025-04-12
Fails to parse query containing `%2B` ( urlencoded `+` ) when query mark as array of date/string.


drsmile1001:
### What version of Elysia is running?

1.2.25

### What steps can reproduce the bug?

```ts
import { Elysia, t } from "elysia";

const app = new Elysia().get(
  "/",
  ({ query }) => {
    return query;
  },
  {
    query: t.Object({
      date: t.Array(t.Date()), // Array here!
    }),
  }
);

const url = new URL("http://localhost/");
url.searchParams.append("date", "2025-01-01T01:00:00+08:00");

const result = await app
  .handle(new Request(url.toString()))
  .then((res) => res.json());

console.log(result);
```

### What is the expected behavior?

The array of date should be parsed correctly and the result should be:
```ts
{
  date: [ new Date("2025-01-01T01:00:00+08:00") ]
}
```

### What do you see instead?

The array of date is not parsed correctly and the result is:
```ts
{
  type: "validation",
  on: "query",
  summary: "Property 'date' should be one of: 'ArrayString', 'array'",
  property: "/date",
  message: "Expected union value",
  expected: {},
  found: {
    date: [ "2025-01-01T01:00:00 08:00" ], // notice the missing '+' sign
  },
}
```

### Additional information

I looked into issues related to encoding `+`, including:

- #821 Fixed!
- #924 It seems not a bug, but not answered yet.
- #943 I think it was fixed after 1.2.15 to now (1.2.25).
- #1001 Like #943, I think it was fixed after 1.2.15 to now (1.2.25).

I also saw an explanation of the behavior of URLSearchParams: https://github.com/elysiajs/elysia/pull/960#issuecomment-2661362653.  
I also believe following this behavior is correct. Therefore, the example code above generates the query string using URLSearchParams.

However, it seems that the Elysia parsing middleware still has some issues when dealing with arrays. If not wrapped in an array, the current behavior is correct.

Edited, give a sample to show current issue is about array of string has problem.
```ts
import Elysia, { t } from "elysia";

const api = new Elysia().get("", ({ query }) => query, {
  query: t.Object({
    array: t.Array(t.String()), // Mark as an array
    simple: t.String(), // Mark as not an array
  }),
});

const url = new URL("http://localhost/");
url.searchParams.append("array", "a+b");
url.searchParams.append("array", "c+d");
url.searchParams.append("simple", "e+f");
[...]

PR #1163, opened 2025-04-14
fix: preserve encoded '+' characters in URL query parameters

SystemVll:
This is a fix for: #1162

## Problem
When processing URL query parameters containing '+' characters (like in dates or mathematical expressions), they were being incorrectly converted to spaces according to standard URL encoding rules. This caused issues with date parameters and other values where literal '+' characters are meaningful.

## Solution
Modified the query parameter handling to preserve encoded '+' and replace not encoded '+' by a space.

## Impact
- Date parameters with time zone indicators (e.g., "2023-04-05T12:30:00+01:00") now parse correctly
- Mathematical expressions containing '+' operators remain intact
- Any query parameter requiring literal '+' characters now works as expected

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -898,12 +898,12 @@ export const composeHandler = ({
 							`if(a${index}===undefined)` +
 							`a${index}=[]\n` +
 							`if(memory===-1){` +
-							`const temp=decodeURIComponent(url.slice(start)).replace(/\\+/g,' ')\n` +
+							`const temp=decodeURIComponent(url.slice(start))\n` +
 							`if(temp.includes(',')){a${index}=a${index}.concat(temp.split(','))}` +
-							`else{a${index}.push(decodeURIComponent(url.slice(start)).replace(/\\+/g,' '))}\n` +
+							`else{a${index}.push(decodeURIComponent(url.slice(start)))}\n` +
 							`break` +
 							`}else{` +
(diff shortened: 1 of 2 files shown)


SaltyAom:
Thanks for the pull request, but I would seconded this.

According to [HTML 4.01 specification](https://www.w3.org/TR/1999/PR-html40-19990824/html40.pdf), section 17.13.4.

> application/x-www-form-urlencoded
This is the default content type. Forms submitted with this content type must be
encoded as follows:
> 1. Control names and values are escaped. Space characters are replaced by ‘+’,
and then reserved characters are escaped as described in [RFC1738]

By default, an HTML form is submitted in an `application/x-www-form-urlencoded` format, which appends data to query parameters for `GET` requests.

This means this PR would break an HTML form parsing functionality.

Although [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-2.1) states that pathname should use `%20` for space but leave an interpretation for query parameters (if I understood it correctly)

Furthermore, according to the Web Standard implementation of [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which treats `+` as space.
```js
console.log(new URL("https://saltyaom.com?message=hello+world").searchParams.get("message"))
// hello world
```

This implied that JavaScript developers would likely expect the implementation of `URLSearchParams` to be the standard.

So I'm likely going to stick with the current implementation of treating `+` as a space unless there's a solid counter-argument for my reasoning.

PR #1165, opened 2025-04-16
fix: skip non-numeric status codes in response schema validation

### LINKED ISSUE #1164: Setting reference model as default response throws an internal server error

1.2.25

Darwin 24.4.0 arm64 arm

When you create a reference model and set it as default response in a route, elysia throws `undefined is not an object (evaluating 'maybeNameOrSchema.type')`.

```tsx
import Elysia, { t } from 'elysia';

const MessageSchema = t.Object({
  message: t.String(),
});

export const appService = new Elysia({
  name: 'appService',
})
  .model({
    message: MessageSchema,
  }, {
    $id: 'message'
  })
  .get(
    '/health',
    () => {
      return {
        timestamp: new Date().toISOString(),
        status: 'ok',
      };
    },
    {
      response: {
        200: t.Object({
          status: t.String(),
          timestamp: t.String(),
        }),
        default: t.Ref('message')
      }
    },
  )
  .listen(3000);
```

It should return 200 as response.

### What do you see instead?

It returns `undefined is not an object (evaluating 'maybeNameOrSchema.type')` as response. The error doesn't even reach `onError` hook.

### Additional information

Even setting value of default as `message` without `t.Ref` returns the same error.

yes

ajaykarthikr:
When defining response schemas with status codes, only numeric HTTP status codes (like 200, 404, etc.) should be processed. Previously, the validator would crash if non-numeric keys were present in the response schema definition.

Error
```
undefined is not an object (evaluating 'maybeNameOrSchema.type')
```

This change adds a validation check to safely ignore any non-numeric status codes, making the schema validator more robust against invalid inputs.

And one of the use case is that we can set a `default` status in response schema as allowed by [OpenAPI 3.0.4](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.4.md#responses-object) specification. 

```tsx
const app = new Elysia().get(
  '/health',
  () => {
    return {
      status: 200,
      response: 'Hello world',
    };
  },
  {
    response: {
      200: t.String(),
      default: t.String(),
    },
  },
);
```

This will be generate in Swagger UI like this:

Screenshot 2025-04-16 at 3 27 44 PM

 

Also, fixes #1164

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -844,6 +844,10 @@ export const getResponseSchemaValidator = (
 	const record: Record<number, TypeCheck<any>> = {}
 
 	Object.keys(maybeSchemaOrRecord).forEach((status): TSchema | undefined => {
+		if (isNaN(+status)) {
+			return undefined;
+		}
+		
 		const maybeNameOrSchema = maybeSchemaOrRecord[+status]
 
 		if (typeof maybeNameOrSchema === 'string') {
--- a/test/validator/response.test.ts
+++ b/test/validator/response.test.ts
@@ -257,7 +257,8 @@ describe('Response Validator', () => {
 				}),
 				response: {
 					200: t.String(),
-					201: t.Number()
+					201: t.Number(),
+					default: t.String()
 				}
 			}
 		)


Inplex-sys (approved):

SaltyAom:
Merged to #1141 via [85c28b3](https://github.com/elysiajs/elysia/commit/85c28b3a523587aae1a20b2bd1c16631b320e393)

Issue #1212, opened 2025-05-14
elysia@1.3 broke node adapter


fapspirit:
### What version of Elysia is running?

1.3.1

### What platform is your computer?

Darwin 24.0.0 arm64 arm

### What steps can reproduce the bug?

- Install `elysia@1.3`
- Install `@elysiajs/node@0.0.3`
- Run a basic server:
```ts
import { Elysia } from "elysia";
import { node } from "@elysiajs/node";

new Elysia({ adapter: node() })
    .get("/", () => "Hello Elysia")
    .listen(3000, ({ hostname, port }) => {
        console.log(`🦊 Elysia is running at ${hostname}:${port}`);
    });
```

### What is the expected behavior?

Server is successfully started, console output is:
```
🦊 Elysia is running at :::3000
```

### What do you see instead?

```js
import { getSchemaValidator, isNotEmpty as isNotEmpty2, randomId } from "elysia/utils";
         ^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module 'elysia/utils' does not provide an export named 'getSchemaValidator'
```

### Additional information

Related issue in node adapter repo: https://github.com/elysiajs/node/issues/40

The original error is easy to fix since `getSchemaValidator` was moved from "elysia/utils" to "elysia/schema", but once fixed, another error come up which is not trivial to fix:
```js
import { mergeResponseWithSetHeaders } from "elysia/adapter/web-standard/handler";
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: The requested module 'elysia/adapter/web-standard/handler' does not provide an export named 'mergeResponseWithSetHeaders'
```

Looks like this file was removed in this commit: https://github.com/elysiajs/elysia/commit/1d2c6627b8d49b9a1fae2c8df34105c607f1d3a1 and it's not obvious from function names what is the replacement for this function.

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

kravetsone:
Same issue

wrong7:
Same

alaarihan:
+1

fapspirit:
Looks like the problem has been fixed on the node adapter (starting `@elysiajs/node@1.3`) and everything is working as expected, so I'll close this issue.

PR #1216, opened 2025-05-20
export adapter utils for use in building external adapters

bobjohnbob:
This PR will add the /adapter/utils file to the list of exports in package.json.

Currently the node adapter is [broken](https://github.com/elysiajs/elysia/issues/1212) on Elysia 1.3.  Some of us have attempted to fix it, but it seems that having access to import the adapter utils will be necessary to do so without copying or rebuilding a bunch of code.  Exporting the these utils in "package.json should allow the node adapter to access those utils and reuse the code that is used to build the built-in adapter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,6 +108,11 @@
 			"import": "./dist/adapter/index.mjs",
 			"require": "./dist/cjs/adapter/index.js"
 		},
+		"./adapter/utils": {
+			"types": "./dist/adapter/utils.d.ts",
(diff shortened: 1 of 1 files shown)


jakub-msqt:
Thank you for the fix. Hopefully this gets merged ASAP, I'm blocked on this.

SaltyAom:
Close as fixed by 47ae5fe30112d3321c4189e993c2a975292ea0f0

PR #1222, opened 2025-05-25
Fixed an issue with memoirist not being a runtime dependency

### LINKED ISSUE #1221: Memoirist should be a runtime dependency

1.3.1

_No response_

Bundle an app using elysia using `bun build`

The app should build.

### What do you see instead?

Error:
```
error: Could not resolve: "memoirist". Maybe you need to "bun install"?
    at [...]/node_modules/elysia/dist/bun/index.js:2:553
```

### Additional information

Offending line:

https://github.com/elysiajs/elysia/blob/9ba8551cecd90034856fa05833b6c4985a3fb6cf/src/index.ts#L1

Looks like the commit https://github.com/elysiajs/elysia/commit/7e2f35bea4b51ce6ac26aa8dddd21f35e42ac8ba moved it into `devDependencies` despite the fact it is used in 
https://github.com/elysiajs/elysia/blob/9ba8551cecd90034856fa05833b6c4985a3fb6cf/src/index.ts#L276-L284.

yes

RafalGoslawski:
Fixes #1221.

### DIFF
--- a/package.json
+++ b/package.json
@@ -180,7 +180,8 @@
 	"dependencies": {
 		"cookie": "^1.0.2",
 		"exact-mirror": "0.1.2",
-		"fast-decode-uri-component": "^1.0.1"
+		"fast-decode-uri-component": "^1.0.1",
+		"memoirist": "^0.3.0"
 	},
 	"devDependencies": {
 		"@types/bun": "^1.2.14",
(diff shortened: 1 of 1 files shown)


SaltyAom:
Elysia is using tsup and Bun build

Both will bundle `memoirist` directly to code if specified as `devDependencies`

PR #1241, opened 2025-06-04
fix: standalone validator - guard types, as(scoped|global) and hooks

Lolotronop:
As of now, standalone validators do not correctly work with:
- `.as("scoped" | "global")` on the instance (standaloneValidators property doesn't get promoted to the proper level)
- hooks inside the guard like `beforeHandle` don't get set on the instance
- `.derive`, `.onBeforeHandle` and most other hooks (type information from standalone validators doesn't get to them)

This PR aims to fix this. Standalone validators are the best feature imho as it enables great composability in the framework, but they are a bit undercooked. Thanks for creating and maintaning elysia!

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2902,7 +2902,6 @@ export default class Elysia<
 			)
 			this.validator.local = null
 
-			console.log("BEFORE", this.standaloneValidator.local)
 			if (this.standaloneValidator.local !== null) {
 				this.standaloneValidator.scoped ||= []
 				this.standaloneValidator.scoped.push(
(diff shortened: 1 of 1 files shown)


dy0gu:
@SaltyAom can you take a look at this? It would fix issues with header schemas of different instance routes not merging!

Lolotronop:
Again, this does not currently affect elysia-swagger, but it will make the experience better once support for standaloneValidators lands there. Also I don't think that pinging the maintainer is helpful. They will get to it when they can, respect their free time. If they don't, that means this PR is not what this project needs and I'm not the one to argue it

Also, this PR doesn't actually merge any internal structures. standaloneValidators are, as the name suggests, implemented as an array of validators that have to be applied sequentially, living alongside the regular ("merged", tho a better word would be overriden) validators on the route. Only the types are merged for the type checking to work. For the OpenAPI generation, however, they would need to be merged in runtime-land. That is something that should be done in the openapi-swagger package tho

PR #1250, opened 2025-06-09
fix: align mapResponse and onAfterHandle with unified return types an…

slowcode42:
### Changes

- Relaxed the return type of `mapResponse` to support primitive values (e.g. string, number), aligning its behavior with `AfterHandler`.
- Updated `onAfterHandle` to use `AfterHandler` instead of `OptionalHandler` for consistency and clarity in handler semantics.

### Motivation

This change improves the developer experience by making the behavior of response handlers more predictable and flexible. It also eliminates subtle inconsistencies between `mapResponse` and `onAfterHandle`.

No breaking changes expected for users following the current type contracts.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1951,7 +1951,7 @@ export default class Elysia<
 	>(
 		options: { as?: LifeCycleType },
 		handler: MaybeArray<
-			OptionalHandler<
+			AfterHandler<
 				MergeSchema<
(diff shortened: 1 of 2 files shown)


SaltyAom:
Unfortunately, I don't think this will work

As `mapResponse` is expected to return a `Response`

PR #1313, opened 2025-07-29
fix: onRequest not firing

### LINKED ISSUE #1308: CORS origin is broken in v1.3.6

1.3.6

MacOS

origin property in the CORS plugin completely ignores any input. 

Leads to all requests being ignored

This case is the only one that I tested.
Array<string | RegExp | Function> - iterate through all cases above in order, allowed if any of the values are true.

Allowed origins should be configurable

### What do you see instead?

localhost:3000 - 3003 where not able to talk to my backend even though it has been working in previous versions. Solution was to revert to 1.3.5.

### Additional information

_No response_

Yes. Did not fix.

rkaw92:
This fixes an issue where `onRequest` handlers would not fire, breaking plugins such as `@elysiajs/cors`.

I believe this was caused by not including `createOnRequestHandler()` in the compose phase, even if some `request` handlers should be registered.

Fixes #1308


Albertbol:
No tests?

Issue #1353, opened 2025-08-22
DANGEROUS! Serialization of objects in array leaks extra fields not included in response model


eyueldk:
### What version of Elysia is running?

elysia@1.3.15

### What platform is your computer?

Microsoft Windows NT 10.0.19045.0 x64 & Linux 6.1.43 x86_64 unknown

### What environment are you using

Bun 1.1.38

### Are you using dynamic mode?

no

### What steps can reproduce the bug?

```typescript
import { Elysia, t } from "elysia";
import { swagger } from "@elysiajs/swagger";

const UserRead = t.Object({
  username: t.String(),
});

const app = new Elysia()
  .use(swagger())
  .get(
    "/",
    () => {
      return {
        hasMore: true,
        total: 1,
        offset: 0,
        totalPages: 1,
        currentPage: 1,
        items: [{ username: "Bob", secret: "shhh" }],
      };
    },
    {
      response: {
        200: t.Object({
          hasMore: t.Boolean(),
          items: t.Array(UserRead),
          total: t.Integer({ minimum: 0 }),
          offset: t.Integer({ minimum: 0 }),
          limit: t.Optional(t.Integer({ minimum: 1 })),
          totalPages: t.Integer({ minimum: 1 }),
          currentPage: t.Integer({ minimum: 1 }),
        }),
      },
    }
  )
  .listen(3000);

console.log(
  `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}/swagger`
);
```

### What is the expected behavior?

Only fields specified in Response model should be sent. i.e. only username is exposed.

### What do you see instead?

Extra fields not specified in response model is sent, i.e. secret is exposed.

### Additional information

The weirdest thing is, if you remove hasMore from response model, it works as expected.
If you specify `additionalProperties: false` in `UserRead` it works (this should be default behavior)
Using `typebox` `Value.Parse` also works as expected.
For some reason this exact model breaks `Elysia`
TEST: You may try my code @ https://codesandbox.io/p/devbox/nlk56y

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

mmorasch:
this code also leaks the secret:
```typescript
import { Elysia, t } from "elysia";

const UserRead = t.Object({
  username: t.String(),
});

const app = new Elysia()
  .get(
    "/",
    () => {
      return {
        total: 1,
        offset: 0,
        totalPages: 1,
        currentPage: 1,
[...]

PR #1355, opened 2025-08-22
fix: prevent response field leakage in API responses (fix for #1353 )

mmorasch:
## Summary
This PR fixes a security vulnerability where extra fields not defined in response schemas were being leaked in API responses.

## Root Cause
The issue occurred due to two main problems:
1. The `hasAdditionalProperties` function incorrectly returned early with undefined values, preventing proper schema cleaning
2. The AOT compilation only applied Clean functions during error recovery, not during normal response processing

## Changes
- Remove problematic early return in `hasAdditionalProperties` function
- Improve `additionalProperties` condition check to handle undefined values  
- Modify response validation to apply Clean function before encoding/validation
- Add comprehensive test suite covering various field leak scenarios

## Test Results
- ✅ All 7 new test cases pass
- ✅ All existing validator tests pass (184/184)
- ✅ All existing response tests pass (34/34) 
- ✅ No regressions detected

## Security Impact
This fix prevents sensitive data from accidentally leaking in API responses when using response schemas for validation.

## Test Plan
- [x] Created comprehensive test cases covering nested objects, arrays, and various schema configurations
- [x] Verified fix works in both dynamic and AOT compilation modes
- [x] Confirmed no regressions in existing functionality

## Summary by CodeRabbit

- Bug Fixes
  - Prevents API responses from leaking fields not defined in the schema, including within nested objects and arrays.
  - Respects additionalProperties: true to preserve explicitly allowed extra fields; defaults to disallowing extras when undefined.
  - Adjusts validation flow to avoid duplicate cleaning and apply cleaning at the correct time.

- Tests
  - Added comprehensive tests ensuring non-schema fields are stripped across nested structures, arrays, mixed/optional schemas, and a reproduced leak scenario.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -256,7 +256,12 @@ const composeValidationFactory = ({
 					ignoreTryCatch
 				})
 
-			if (appliedCleaner) code += clean()
+			// Always apply Clean first if we have normalize and Clean function
+			if (!appliedCleaner && normalize && value.Clean) {
+				code += clean()
+			} else if (appliedCleaner) {
+				code += clean()
+			}
 
 			const applyErrorCleaner =
 				!appliedCleaner && normalize && !noValidate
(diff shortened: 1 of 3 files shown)


SaltyAom:
Similar to [9c68be6](https://github.com/elysiajs/elysia/commit/9c68be6307a817f2b1bb5a9dcd1d662683c3372b)

Closing as complete

Issue #260, opened 2023-10-08
How to connect socket.io?


heitoke:

cool3rain:
socket.io does work with the ElysiaJS which is based on Bun which uses uWebSockets so there's no need to use socket.io

heitoke:
Will I have my own methods for accepting and sending data? And how do I connect from the client (socket.io-client) to the server?

cool3rain:
`
// server.tsx
import { Elysia } from "elysia";
import { cors } from '@elysiajs/cors';

const app = new Elysia()
  .use(cors())
  .ws('/ws', {
      open: (ws) => {
        console.log(ws.data);
      },
      message: (ws, message) => {
        console.log(message);
      },
      close: (ws) => {
        console.log(ws)
      }
    })
    .listen(8080)

console.log(
  `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
);



// client.tsx
const WebSocket = require("ws");

const ws = new WebSocket("http://localhost:8080/ws");

ws.on("error", console.error);

ws.on("open", function open() {
  ws.send("your_data");
});

ws.on("message", function message(data) {
  console.log("received: %s", data);
});`

Issue #162, opened 2023-09-16
Socket.io integration


arsaizdihar:
Currently, bun support socket.io https://github.com/oven-sh/bun/discussions/2111

I would love to see at least the documentation about how to integrate socket.io with Elysia (or even better the plugin)

alevel7:
have you checked this out https://elysiajs.com/patterns/websocket.html

SaltyAom:
Closing with [WebSocket](https://elysiajs.com/patterns/websocket.html) and [Eden WS](https://elysiajs.com/plugins/eden/treaty.html#websocket)

gurachan:
> Currently, bun support socket.io [oven-sh/bun#2111](https://github.com/oven-sh/bun/discussions/2111)
> 
> I would love to see at least the documentation about how to integrate socket.io with Elysia (or even better the plugin)

Same like  ```io.attachApp(elysiaApp)```
the only reason is so the server http/https will be handle by elysia and socket.io can still do its thing.. also socket.io uses ws by default.. means it uses uwebsocket too if u run it on bun.. 

for now im using two ports one for server(elysia) one for ws(socket.io) I dont like reinventing the wheel theres some stuff in socket.io that i have to code it my self if i move on other ws xD i mean both of them will use the uwebsocket in bun anyway.

Wafje:
> > Currently, bun support socket.io [oven-sh/bun#2111](https://github.com/oven-sh/bun/discussions/2111)
> > I would love to see at least the documentation about how to integrate socket.io with Elysia (or even better the plugin)
> 
> Same like `io.attachApp(elysiaApp)` the only reason is so the server http/https will be handle by elysia and socket.io can still do its thing.. also socket.io uses ws by default.. means it uses uwebsocket too if u run it on bun..
> 
> for now im using two ports one for server(elysia) one for ws(socket.io) I dont like reinventing the wheel theres some stuff in socket.io that i have to code it my self if i move on other ws integration xD i mean both of them will use the uwebsocket in bun anyway.

Hi @gurachan,
How did you set this up? Can you share some code snippets?
I can't seem to be able to connect to my app (with socket.io).

FYI: Requiring Socket.IO for acknowledgement functionality.

PR #1358, opened 2025-08-25
fix: allow overriding websocket handler with listen options

gkurt:
Many people have asked for Socket.io integration at #260 and #162 and it's finally here. Socket.io added support for Bun. It can be seen in [this post](https://github.com/oven-sh/bun/discussions/2111#discussioncomment-14186172).

But to make integration easier with Elysia, the websocket handler must be configurable. This PR makes a small tweak to the `listen` function to allow that.

After this, it will be possible to add Socket.io to Elysia like this:

```ts
import { Server as Engine } from '@socket.io/bun-engine';
import { Server } from 'socket.io';

const io = new Server();
const engine = new Engine({ path: '/socket.io/' });
io.bind(engine);

const app = new Elysia()
  .all('/socket.io/', ({ request, server }) => engine.handleRequest(request, server!))
  .listen({
    idleTimeout: 30,
    websocket: engine.handler().websocket as unknown as Bun.WebSocketHandler,
  });
```

Note that adding socket.io like this will disable Elysia's websocket features.

## Summary by CodeRabbit

- New Features
  - You can now pass websocket configuration via the options object when starting the Bun server, enabling customization of WebSocket behavior (e.g., timeouts, compression, limits).
- Refactor
  - Improved merging of WebSocket options to prioritize user-specified settings provided in the options object.
- Chores
  - No changes to public API signatures; existing setups continue to work without modification.

### DIFF
--- a/src/adapter/bun/index.ts
+++ b/src/adapter/bun/index.ts
@@ -341,7 +341,9 @@ export const BunAdapter: ElysiaAdapter = {
 							),
 							websocket: {
 								...(app.config.websocket || {}),
-								...(websocket || {})
+								...(websocket || {}),
+								// @ts-expect-error not available in this variant of options type
+								...(options.websocket || {})
 							},
 							fetch: app.fetch
 						} as Serve)


gkurt:
@SaltyAom can you review this if it's not too much trouble?

SaltyAom:
Thanks for the PR!

This PR basically fixes the websocket argument not being passed to Bun serve, which is technically my fault.

Technically, this could already be implemented with a constructor argument like this:
```ts
new Elysia({
	websocket: engine.handler().websocket as unknown
})
```

SaltyAom:
Don't worry, the unit tests actually passed, but failed because running on Bun 1.2.21 but missing #1365

gkurt:
Hmm, I tried both `new Elysia({ websocket: ... })` and `new Elysia({ server: { websocket: ... } })` and both don't work. Only `app.listen({ websocket: ... })` works after this PR. I guess Elysia overrides websocket config in the other methods.

Issue #20, opened 2023-02-18
typia/zod support


gotjoshua:
any chance?

SaltyAom:
I have this consideration I have previously with Zod/io-ts/typia.

For one, I want to put out as fast as possible while still balancing with Quality of Life, Typia introduce has 2 mature problems that make me not support Typia (at least for now).

1. It shouldn't only just check the type, the reason I pick TypeBox is that it also supports OpenAPI Schema which is important to make a standard production-grade REST API server, and the ability to specify server specs and generated documentation of fly.

> Many have asked why Elysia uses type box instead.
> 1. Typebox is built around OpenAPI which made the Swagger plug-in possible.

2. Typia requires an additional setup step and forces the developer to go TypeScript-only.

3. The cost of adding an extra union type to support End-to-End type safety is likely very high and complex, for one to make this full type-safety work, Elysia is using depends on [LocalHandler](https://github.com/elysiajs/elysia/blob/main/src/types.ts#L352), [LocalHook](https://github.com/elysiajs/elysia/blob/main/src/types.ts#L352), [ElysiaRoute](https://github.com/elysiajs/elysia/blob/main/src/types.ts#L269) [TypeRouteToEden](https://github.com/elysiajs/elysia/blob/main/src/types.ts#L299), and [CreateEden](https://github.com/elysiajs/eden/blob/main/src/types.ts#L48) which is highly a complex type consist of over 600 loc for typing only, at the moment the type inference isn't as fast as it should be but if we manage to add one more union type for Typia, it's likely to have a problem with the performance of type inference.

And in regard to performance, seems like TypeBox is experimenting with [AoT compile to improve performance](http://sinclairzx81.github.io/runtime-type-benchmarks)

## Summary
We may support Typia someday or might not support Typia at all (depending on the situation that may on fold in the future), but for now we have no interest or priority to support Typia at moment, and if we have to support 2nd validation library one day, it's likely that Zod will be our priority next.

Will leave this issues open for a week, feels free to ask more question if the answer is not cleared or if you have more questions want to ask, feels free.

gotjoshua:
very satisfying answer, thanks @SaltyAom !

I also wasn't thrilled with the ttsc business and very much enjoyed my approach to typebox and zod.

thanks for all your work and generosity!

murraybauer:
[...]

PR #1381, opened 2025-09-05
Standard Schema

SaltyAom:
Addressing #20 by adding support for Standard Schema

- [x] validate successfully
- [x] it infers type correctly
- [x] work with treaty (only proper type inference is required)
- [x] work as a standalone schema
- [ ] work as reference schema (string name)
- [ ] has test cases
- [x] work with dynamic mode (it's likely to work now, but not test yet)
- [ ] support OpenAPI by library's native API annotation (planned, likely will use [xs schema])(https://www.npmjs.com/package/xsschema)
- [x] support with OpenAPI for a library with no support for OpenAPI somehow. Solved with [Elysia OpenAPI Type Gen](https://elysiajs.com/blog/openapi-type-gen.html)
- [x] type inference is fast enough that it doesn't have a significant slowdown
- [x] support multiple schemas from multiple libraries interchangeably
- [x] it normalizes when using multiple standalone schema⁽¹⁾

1. Works by creating a parsed snapshot value of each validator, then merging them together. See https://github.com/elysiajs/elysia/commit/a43a60790a7e19000d28b9418b6c9d6c5146decd.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,9 +4,11 @@
     "": {
       "name": "elysia",
       "dependencies": {
+        "arktype": "^2.1.20",
         "cookie": "^1.0.2",
         "exact-mirror": "0.1.6",
         "fast-decode-uri-component": "^1.0.1",
+        "valibot": "^1.1.0",
       },
       "devDependencies": {
         "@types/bun": "^1.2.16",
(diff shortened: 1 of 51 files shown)


zvictor:
Wonderful! Elysia should now be listed at https://github.com/standard-schema/standard-schema?tab=readme-ov-file#what-tools--frameworks-accept-spec-compliant-schemas

Issue #1395, opened 2025-09-10
MaybeEmpty causes type to be inferred as unknown


KurtGokhan:
### What version of Elysia is running?

1.3.21

### What platform is your computer?

Darwin 24.6.0 arm64 arm

### What environment are you using

Bun v1.2.21

### Are you using dynamic mode?

No

### What steps can reproduce the bug?

When using `t.MaybeEmpty`, the top-level type is inferred as `unknown`. The issue doesn't seem to happen with `t.Nullable` etc. But it's also reproducible when doing an union with undefined, e.g. `t.Union([t.Undefined(), ...])`

<img width="1164" height="378" alt="Image" src="https://github.com/user-attachments/assets/2d3d1050-3294-4f7d-8891-9a45ec95d089" />

### What is the expected behavior?

Expected type of query to be `{ name: string } | undefined`

### What do you see instead?

Type of query is `unknown`

### Additional information

This code can reproduce the issue

```ts
import { expectTypeOf } from "bun:test";
import { Elysia, t } from "elysia";

const app = new Elysia().get("/", () => "Hello Elysia", {
  query: t.MaybeEmpty(t.Object({ name: t.String() }))
});

expectTypeOf(app["~Routes"].get.query).not.toEqualTypeOf<unknown>();
```

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

Yes

PR #1397, opened 2025-09-11
fix: infer type of undefinable parameters correctly

gkurt:
Addresses #1395

The main cause of the problem is, `undefined extends T` is true for `T = undefined | Something` so any type that had that signature was being resolved to the default type.

I added proper `IsUndefined` check. Also noticed that it should also fallback to the default type when the type is unknown, so I added `IsUnknown` check. Then, I merged them in the `IsSkipped` check so that can be used generally for these cases. If the condition to fallback to the default type changes in the future, it will be enough to change the implementation of `IsSkipped`.

Finally, added a `t.Undefinable` helper. `t.Optional` would have been a better name but Typebox uses it for making a schema partial. `t.MaybeEmpty` also includes `null` type, and that makes this syntax not possible:

```ts
new Elysia().post("/", ({ query: { name = 'Elysia' } = {} }) => `Hello ${name}`, {
  query: t.Undefinable(t.Object({ name: t.String() })),
});
```

I know the changes I made are somewhat risky and intrusive. And I know you are working on integrating standard-schema. So feel free to modify or discard this PR. But I think a change like this is needed, as there is no way to have a type that is unioned with undefined at the moment.

Note: A similar change is also needed for Eden, and I am planning to also make that if this is merged. Because the same problem exists there too.

## Summary by CodeRabbit

- New Features
  - Added t.Undefinable to define optional (undefined-only) values without allowing null across schemas.

- Improvements
  - More accurate TypeScript inference for query, params, headers, and cookie in route contexts, especially when segments are optional or skipped.
  - Enhanced schema merging behavior to better respect optional/unknown inputs for body, headers, query, and cookie.

- Tests
  - Added type-level tests validating t.Undefinable inference for route inputs (query, headers, body) on POST endpoints.

### DIFF
--- a/src/context.ts
+++ b/src/context.ts
@@ -12,7 +12,8 @@ import type {
 	Prettify,
 	ResolvePath,
 	SingletonBase,
-	HTTPHeaders
+	HTTPHeaders,
+	IsSkipped
 } from './types'
 
 type InvertedStatusMapKey = keyof InvertedStatusMap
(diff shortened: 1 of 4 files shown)


SaltyAom:
> `t.Optional would have been a better name but Typebox uses it for making a schema partial.`

In Elysia, Optional can also make a field optional as well
See https://elysiajs.com/patterns/typebox.html#optional-1

gkurt:
Sorry, this is outdated after the changes in 1.4. I will try to update this. I believe the problem in Elysia might be fixed already, but Eden still doesn't handle it well.

gkurt:
@SaltyAom  not sure if I understand well. The example you show produces a type like this:

```
query: {
    name?: string | undefined;
}
```

With `Undefinable`, it produces a type like this:

```
query?: {
    name?: string | undefined;
}
```

So it's not the same. This can be done in other schemas, e.g. Zod, but not Typebox the last time I checked.

Issue #1458, opened 2025-10-05
`strictPath` is forced to true when `aot: false` is set


Chipsnet:
### What version of Elysia is running?

elysia@1.4.9

### What platform is your computer?

Darwin 24.6.0 arm64 arm

### What environment are you using

Bun 1.2.20

### Are you using dynamic mode?

Yes

### What steps can reproduce the bug?

Set `aot: false` like this:

```ts
import { Elysia } from "elysia";

const app = new Elysia({
  aot: false,
})
  .get("/", () => "Hello Elysia")
  .get("/ping", () => "pong")
  .group("/api", (app) =>
    app.get("/", () => "Hello Elysia").get("/ping", () => "pong"),
  )
  .listen(3000);

console.log(
  `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`,
);
```

Then send GET requests to the following URLs:

✅ http://localhost:3000/  
✅ http://localhost:3000/ping  
❌ http://localhost:3000/ping/  
❌ http://localhost:3000/api  
✅ http://localhost:3000/api/  
✅ http://localhost:3000/api/ping  
❌ http://localhost:3000/api/ping/  

✅: 200 / ❌: 404

### What is the expected behavior?

When `aot: false` is set and `strictPath` is not specified (thus defaulting to `false`),  
the following requests should succeed:

✅ http://localhost:3000/ping/  
✅ http://localhost:3000/api  
✅ http://localhost:3000/api/ping/  

In other words, the behavior should match the default `strictPath: false` mode,  
where both paths—with and without trailing slashes—are treated as equivalent.

### What do you see instead?

The routes behave **as if `strictPath: true` were enabled**,  
even though it is not explicitly set.

- Paths with trailing slashes (e.g., `/ping/`, `/api/ping/`) return **404 Not Found**.  
- Paths without trailing slashes (e.g., `/ping`, `/api/ping`) work correctly.  
- The same behavior occurs for nested routes under `.group()` as well.

If `strictPath` is explicitly set to `false`, it works correctly as expected.  
However, `strictPath` defaults to `false`, and this issue only occurs when `aot: false` is set.  
When `aot` is not set (default: `true`), routing behaves correctly with `strictPath: false`.

So, setting `aot: false` seems to unintentionally make the app behave as if `strictPath` were `true`.

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

yes

PR #1459, opened 2025-10-05
Fix/aot strictpath default

MarcelOlsen:
## Description

Fixes strictPath default behavior when `aot: false` is set. This solves #1458 

## The Bug

When you set `aot: false` without explicitly specifying `strictPath`,
routes behave as if `strictPath: true` is enabled. This causes paths with
trailing slashes to return 404.

**Example:**
```typescript
const app = new Elysia({ aot: false })
  .get("/ping", () => "pong")
  .group("/api", (app) => app.get("/ping", () => "pong"))
```

#### Returns 404, should be 200
```typescript
await app.handle(req("/ping/"))
await app.handle(req("/api/ping/"))
```

## Root Cause

Line 812 in src/index.ts uses strict equality:
```typescript
if (this.config.strictPath === false) {
```

  When strictPath is undefined (the default), undefined === false evaluates
  to false, so loose path variants never get added to the dynamic router.

  The Fix

  Changed to truthy check:
  ```typescript
  if (!this.config.strictPath) {
  ```

  This matches the pattern used elsewhere in the codebase (lines 1024, 1042,
   1048, and compose.ts:2271).

  Tests

  Added test/core/aot-strictpath.test.ts covering:
  - Default strictPath behavior with aot:false
  - Explicit strictPath:false with aot:false
  - strictPath:true with aot:false
  - Group routes with default strictPath


MarcelOlsen:
@coderabbitai review

PR #1464, opened 2025-10-09
Allow using static handlers in Bun v1.3

Jarred-Sumner:
The version check for `supportPerMethodInlineHandler` started returning false when bumping the version of Bun from Bun v1.2 -> Bun v1.3

This led to [`test/core/native-static.test.ts`](https://github.com/elysiajs/elysia/blob/817a4656cf3330c98f23e425908ca2b090894a46/test/core/native-static.test.ts) failing

image

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with newer Bun runtimes (>=1.2.14) by optimizing version detection, preventing incorrect behavior in certain environments while leaving older versions unaffected.
  * Streamlined runtime checks to avoid unnecessary parsing, reducing overhead and stabilizing behavior during startup.
  * Enhances reliability of feature gating across environments. No public API changes.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1078,6 +1078,8 @@ export const encodePath = (path: string, { dynamic = false } = {}) => {
 export const supportPerMethodInlineHandler = (() => {
 	if (typeof Bun === 'undefined') return true
 
+	if (Bun.semver?.satisfies?.(Bun.version, '>=1.2.14')) return true
+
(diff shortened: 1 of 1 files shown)


SaltyAom:
Oh, that's so silly of me ￼

SaltyAom:
I've published this under 1.4.10, it should be working now. Sorry for the inconvenient

PR #1521, opened 2025-11-01
feat: Add stream autoCancellation configuration option for streams

Aunali321:
This PR adds a configuration option to control automatic cancellation of generator-based streams when clients disconnect.

Changes:
- Added stream config option to ElysiaConfig with autoCancellation flag
- Updated CreateHandlerParameter to accept streamOptions
- Updated createStreamHandler to respect autoCancellation setting
- Added ELYSIA_STREAM_AUTO_CANCELLATION environment variable support
- Updated both Bun and Web Standard handlers to use the new option
- Added comprehensive tests for both enabled and disabled autoCancellation behavior

By default, autoCancellation is enabled (true) to maintain backward compatibility.

Usage:
1. Environment variable: ELYSIA_STREAM_AUTO_CANCELLATION=false
2. Programmatic: pass streamOptions to createStreamHandler

Addresses the use case where developers need generator functions to continue executing even after a client disconnects.

Motivation:

Elysia's automatic stream cancellation on client disconnect prevents valid use cases where server operations should complete regardless of connection status (e.g., AI agent executions that persist results to database, paid API calls that cannot be resumed, background job processing).

This PR adds an opt-in configuration flag to disable auto-cancellation while maintaining backward compatibility, enabling developers to build resilient systems where long-running operations complete even when clients disconnect and reconnect later.

## Release Notes

* **New Features**
  * Added a new stream configuration option enabling automatic cancellation of streaming responses when clients disconnect. This feature is enabled by default, ensuring graceful termination of ongoing streaming operations upon client disconnection.

* **Tests**
  * Added comprehensive tests validating auto-cancellation behavior for streaming responses with the feature both enabled and disabled.

### DIFF
--- a/src/adapter/bun/handler.ts
+++ b/src/adapter/bun/handler.ts
@@ -567,5 +567,6 @@ const handleResponse = createResponseHandler({
 
 const handleStream = createStreamHandler({
 	mapResponse,
-	mapCompactResponse
+	mapCompactResponse,
+	streamOptions: {}
 })
--- a/src/adapter/utils.ts
+++ b/src/adapter/utils.ts
@@ -143,17 +143,22 @@ type CreateHandlerParameter = {
 		request?: Request
 	): Response
 	mapCompactResponse(response: unknown, request?: Request): Response
+	streamOptions?: {
+		autoCancellation?: boolean
+	}
 }
 
 const allowRapidStream = env.ELYSIA_RAPID_STREAM === 'true'
+const defaultAutoCancellation = env.ELYSIA_STREAM_AUTO_CANCELLATION !== 'false'
 
 export const createStreamHandler =
-	({ mapResponse, mapCompactResponse }: CreateHandlerParameter) =>
+	({ mapResponse, mapCompactResponse, streamOptions }: CreateHandlerParameter) =>
 	async (
 		generator: Generator | AsyncGenerator | ReadableStream,
 		set?: Context['set'],
 		request?: Request
 	) => {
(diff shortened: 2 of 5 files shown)


Aunali321:
Hi SaltyAom,
Thanks for your open-source efforts. I came back to Elysia after a long time, and I think it's fair to say that it's been good enough to rival FastAPI for me.
If you think my motivation for this PR is good, then I'll fix the issues reported by CodeRabbit. If not, I'll close the PR. LMK your thoughts.

akim-bow:
Hi.
I think that this is very specific to your particular usecase which should not affect the library.

You can implement very simple wrappers for your usecase like this:

```
function *a() {
  yield 1;
  yield 2;
  yield 3;
}

let activeSSR = 0;

function *wrap(gen) {
  activeSSR += 1;
  yield* gen;
  activeSSR -= 1;
}

const w = wrap(a())

console.log(activeSSR);
for (const data of w) {
  console.log(activeSSR);
}
console.log(activeSSR);

// Stop Elysia based on activeSSR value
```

I personally do not see any benefits from adding this config to Elysia. It wont work very well when there are at least several clients connected to Elysia. Even in your example with AI responses it won't work well. For example let's say we have one remaining client and after we waited 10 seconds for it to disconnect - Elysia received incoming requests from another client. And then we have another client and then first client wanted to ask for more data... and so on. This can effectively prevent Elysia from stopping for a long time. Adding potentially harmful config option is a bad idea imo. Moreover, it could be simply implemented manually as i showed in a snippet above.

P.S. I'm not related to Elysia.

Issue #1520, opened 2025-10-30
Async handlers crash on HEAD request


artemtam:
### What version of Elysia is running?

1.4.13

### What platform is your computer?

Darwin 25.0.0 arm64 arm (reproduces on x64 Linux as well)

### What environment are you using

Bun 1.3.1

### Are you using dynamic mode?

No

### What steps can reproduce the bug?

Have a GET async handler:

```
import { Elysia } from 'elysia';

const app = new Elysia()
  .get('/async', async () => 'response')
  .listen(3000);
```

Make a HEAD request to it:

```
$ curl -I http://localhost:3000/async
```

### What is the expected behavior?

200 OK with a correct Content-Length, [according to the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) – exactly the same as not async handlers work.

### What do you see instead?

```
HTTP/1.1 500 Internal Server Error
```

And an unexpected error in the logs:

```
TypeError: undefined is not an object (evaluating '_res.headers.set')
      at <anonymous> (file:///Users/user/elysia-test/node_modules/elysia/dist/bun/index.js:25:60)
20 | else p=u.substring(s, qi)
21 | const c={request:r,store,qi,path:p,url:u,redirect,status,set:{headers:Object.create(null),status:200}}
22 | 
23 | map: switch(p){
24 | case'/async':case'/async/':switch(r.method){case 'GET':return ht[0].composed(c)
25 | return getResponseLength(_res).then((length)=>{_res.headers.set('content-length', length)
```

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

Yes, set up a clean project to reproduce.

sunnylqm:
@SaltyAom https://github.com/elysiajs/elysia/commit/7937aad26fb1b8b82b8fbfde7250e0894a2f2d84#diff-9043fd343dba2fc691e6d1577d67b4c164b7cdc2ce399d3a3cc787e06dd81c32R2385 breaks it again. typeof promise is not function, but object

PR #1524, opened 2025-11-03
fix head request handler again

sunnylqm:
fix https://github.com/elysiajs/elysia/issues/1520 again

## Summary by CodeRabbit

* **Refactor**
  * Internal restructuring of request handling logic to use Promise-based asynchronous flow instead of conditional branching, improving code maintainability while preserving existing behavior.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -2381,18 +2381,10 @@ export const composeGeneralHandler = (app: AnyElysia) => {
 		)
 			switchMap +=
 				`case 'HEAD':` +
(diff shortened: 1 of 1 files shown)


stewones:
I'm in need of this fix @SaltyAom 

any idea of when we get a new release? maybe it's time to figure out canary builds.

SaltyAom:
@stewones we have patch on going in the "next" branch

The current on going patch is at #1529 which has some PRs pending

As for canary build, you can try

```bash
bun add https://pkg.pr.new/elysiajs/elysia@1529
```

Issue #1540, opened 2025-11-10
toResponse() method ignored if custom error class extends Error


meseta:
### What version of Elysia is running?

1.4.15

### What platform is your computer?

Microsoft Windows NT 10.0.26100.0 x64

### What environment are you using

1.3.1

### Are you using dynamic mode?

no

### What steps can reproduce the bug?

```js
import { Elysia } from "elysia";

class ErrorA extends Error {
  toResponse() {
    return Response.json({ error: "hello" }, { status: 418 });
  }
}

class ErrorB {
  toResponse() {
    return Response.json({ error: "hello" }, { status: 418 });
  }
}

new Elysia()
  .get("/A", () => {
    return new ErrorA()
  })
  .get("/B", () => {
    return new ErrorB()
  })
  .listen(5000);
```

### What is the expected behavior?

GET /A returns status 418 {"error": "hello"}
GET /B returns status 418 {"error": "hello"}

### What do you see instead?

GET /A returns status 500 {"name":"Error","message":""}
GET /B returns status  {"error": "hello"}

### Additional information

According to documentation: https://elysiajs.com/patterns/error-handling.html#custom-error-response, it is possible to create a custom erro response that extends Error, and has a toResponse() method.

However, it appears in the source code, error handling has an `instanceOf Error` case first, before it attempts to handle errors with toResponse(), resulting in toResponse not being used

SaltyAom:
Fixed by #1543 published under 1.4.16

Issue #1539, opened 2025-11-10
toResponse() status not used when thrown, but used when returned


meseta:
### What version of Elysia is running?

1.4.15

### What platform is your computer?

Microsoft Windows NT 10.0.26100.0 x64

### What environment are you using

1.3.1

### Are you using dynamic mode?

no

### What steps can reproduce the bug?

```js
import { Elysia } from "elysia";

class MyError {
  toResponse() {
    return Response.json({ error: "hello" }, { status: 418 });
  }
}

new Elysia()
  .get("/A", () => {
    throw new MyError()
  })
  .get("/B", () => {
    return new MyError()
  })
  .listen(5000);
```

### What is the expected behavior?

GET /A returns status 418
GET /B returns status 418

### What do you see instead?

GET /A returns status 500
GET /B returns status 418

### Additional information

When `throw new MyError()` is used, the response status in toResponse() is ignored, and instead uses a `status` property of the error if one is provided, or 500 otherwise

SaltyAom:
Fixed by #1543 published under 1.4.16

PR #1543, opened 2025-11-11
:wrench: fix: respect toResponse() method on Error classes

MarcelOlsen:
This resolves both #1539 and #1540

## Problem

We had two related bugs with how Elysia handles the `toResponse()` method on error objects:

**Bug 1 (#1540):** When a custom error class extends `Error` and implements `toResponse()`, returning it from a handler would ignore the `toResponse()` method entirely and create a default error response with status 500.

**Bug 2 (#1539):** When throwing an error (Error or non-Error) with `toResponse()`, the method would be called but the status code from the returned Response would be replaced with 500.

Example that was broken:
```typescript
class MyError extends Error {
  toResponse() {
    return Response.json({ error: "custom" }, { status: 418 })
  }
}

// Before: returns status 500 {"name":"Error","message":""}
// After: returns status 418 {"error":"custom"}
app.get('/', () => new MyError())
```

According to the [documentation](https://elysiajs.com/patterns/error-handling.html#custom-error-response), this should work, but the implementation had some gaps.

## Root Cause

**For returned errors:** The `errorToResponse()` function in both adapters was creating default Error responses without checking if the error has a `toResponse()` method first.

**For thrown errors:** The error handler in `compose.ts` was calling `toResponse()`, but the status had already been set to 500 earlier in the catch block. When the Response was passed to `mapResponse()`, it would use the pre-set status instead of extracting it from the Response object.

### 1. Updated errorToResponse() in both adapters
Added a check for `toResponse()` before falling back to default Error response generation. If `toResponse()` exists, we call it and pass the result through `mapResponse()` to handle it consistently with other response types.

**Files changed:**
- `src/adapter/web-standard/handler.ts`
- `src/adapter/bun/handler.ts`

### 2. Updated error handler in compose.ts
Added an early check for `toResponse()` that runs before onError hooks. This check:
- Handles both Error and non-Error objects with `toResponse()`
- Extracts the status code from the Response before passing to `mapResponse()`
- Excludes ValidationError and TransformDecodeError to preserve existing validation error handling
- Takes precedence over onError hooks (the error defines its own response format)

**File changed:**
- `src/compose.ts`

## Design Decisions

**toResponse() takes precedence over onError hooks**

We decided that if an error implements `toResponse()`, it should be used even if there are onError hooks registered. This makes sense because the error class is explicitly defining how it should be represented in the response. If you want onError to override this, you can check for the error type and return your own Response from the hook.

**Validation errors are exempt**

ValidationError and TransformDecodeError are excluded from the toResponse() check to maintain existing behavior for validation errors, which have their own specialized handling.

**Headers and status are both preserved**

The Response returned by `toResponse()` is passed through `mapResponse()`, which means both custom headers and status codes are preserved. This gives full control to the error class over its HTTP response.

## Testing

Added comprehensive test coverage:
- Error classes extending Error with toResponse() (both returned and thrown)
- Non-Error classes with toResponse() (both returned and thrown)
- Custom headers from toResponse()
- All existing 1337 tests still pass

Verified both original bug scenarios now work correctly with the exact code from the bug reports.

## Breaking Changes

None. This fixes behavior to match the documented API, so existing code that wasn't relying on the buggy behavior will continue to work as expected.

## Summary by CodeRabbit

* **New Features**
  * Errors or values that provide a custom response are now honored (sync or async), short-circuiting standard error handling; their responses are mapped and returned.
  * ValidationError and TransformDecodeError are now exposed in the error handler surface.

* **Bug Fixes**
  * Status, headers and redirect info from custom responses are correctly propagated; fallback JSON response preserved for other errors.

* **Tests**
  * Added comprehensive tests covering sync/async, returned vs thrown, Error vs non-Error, headers and status.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,6 +1,6 @@
 import type { AnyElysia } from './index'
 
-import { Value } from '@sinclair/typebox/value'
+import { Value, TransformDecodeError } from '@sinclair/typebox/value'
 import {
 	Kind,
 	OptionalKind,
@@ -2520,6 +2520,8 @@ export const composeErrorHandler = (app: AnyElysia) => {
 		`mapResponse,` +
 		`ERROR_CODE,` +
 		`ElysiaCustomStatusResponse,` +
+		`ValidationError,` +
+		`TransformDecodeError,` +
 		allocateIf(`onError,`, app.event.error) +
 		allocateIf(`afterResponse,`, app.event.afterResponse) +
 		allocateIf(`trace,`, app.event.trace) +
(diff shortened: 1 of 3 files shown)


MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

PR #1549, opened 2025-11-15
🔧  Fix Set-Cookie headers not sent when errors are thrown

### LINKED ISSUE #1547: Set-Cookie not sent to client when Error() on handler

1.4.15

Linux 6.14.0-114036-tuxedo x86_64 x86_64

1.3.2

No

Consider this route:

```js

    .get("/test", async ({cookie}) => {
        cookie.test.value = "a-value";
        throw new Error("test error");
    })

```

The Set-Cookie header is not sent to the client due to the error thrown.

**This happens on response validation erros too**.

The Set-Cookie should be sent anyway.

### What do you see instead?

The Set-Cookie header is not sent to the client due to the error thrown.

### Additional information

For example, I am rotating the session cookie by revoking the previous session id and setting a new session id, but in case of unexpected errors the new session id is not sent to the user and therefore the session is ended.

_No response_

MarcelOlsen:
This resolves #1547

## Problem

Set-Cookie headers are not sent to clients when errors are thrown in route handlers, breaking critical use cases like session rotation on error.

### Current Behavior

```typescript
app.get("/test", async ({ cookie }) => {
    cookie.session.value = "new-session-id"
    throw new Error("unexpected error")
})
```

**Before:**
- Response status: 500 ✓
- Set-Cookie header: **missing** ✗

This affects:
- Session rotation (security-critical)
- Any cookies set before errors
- Both regular errors and validation errors
- Both dynamic and AOT-compiled handlers

### User Impact

When rotating session cookies on errors (e.g., after detecting suspicious activity), the new session ID is never sent to the client. This leaves the user's session in an invalid state, effectively logging them out unexpectedly.

## Root Cause

Error responses were being created directly via `new Response()` without calling `mapResponse()`, which is responsible for serializing cookies from `context.set.cookie` into `Set-Cookie` headers via `handleSet()`.

This happened in two places:

1. **Dynamic error handler** (`src/dynamic-handle.ts:728-734`)
   - Fallback error response created Response directly
   - Cookies in `context.set.cookie` were never serialized

2. **AOT compiled error handler** (`src/adapter/web-standard/index.ts:153-168`)
   - Both `validationError` and `unknownError` handlers created Response directly
   - Applied to both WebStandard and Bun adapters (Bun inherits from WebStandard)

## Solution

Changed all error response paths to use `mapResponse()` instead of creating Response objects directly. This ensures `handleSet()` is called to serialize cookies before the response is created.

### Changes

**1. src/dynamic-handle.ts (lines 728-732)**
```typescript
// Before:
return new Response(
    typeof error.cause === 'string' ? error.cause : error.message,
    {
        headers: context.set.headers as any,
        status: error.status ?? 500
    }
)

// After:
context.set.status = error.status ?? 500
return mapResponse(
    typeof error.cause === 'string' ? error.cause : error.message,
    context.set
)
```

**2. src/adapter/web-standard/index.ts (lines 153-158)**
```typescript
// Before:
validationError:
    `return new Response(` +
    `error.message,` +
    `{` +
    `headers:Object.assign({'content-type':'application/json'},set.headers),` +
    `status:set.status` +
    `}` +
    `)`,
unknownError:
    `return new Response(` +
    `error.message,` +
    `{headers:set.headers,status:error.status??set.status??500}` +
    `)`

// After:
validationError:
    `set.headers['content-type']='application/json';` +
    `return mapResponse(error.message,set)`,
unknownError:
    `set.status=error.status??set.status??500;` +
    `return mapResponse(error.message,set)`
```

### Design Decision

Using `mapResponse()` has the added benefit of preserving **all** headers set via `context.set.headers`, not just cookies. This means CORS headers, custom authentication headers, and other metadata set before an error occurs are also preserved in the error response.

## Testing

Added 9 comprehensive test cases covering:

1. Set-Cookie with regular Error
2. Set-Cookie with response validation errors
3. Set-Cookie with onError hooks
4. Set-Cookie with NotFoundError
5. Set-Cookie with InternalServerError
6. Set-Cookie in AOT mode
7. Multiple cookies preserved
8. Custom headers preserved alongside cookies
9. Headers from set.headers preserved

### Example

After the fix:
```typescript
app.get("/test", async ({ cookie, set }) => {
    cookie.session.value = "new-session-id"
    set.headers['x-request-id'] = 'abc123'
    throw new Error("unexpected error")
})
```

**After:**
- Response status: 500 ✓
- Set-Cookie header: `session=new-session-id; Path=/` ✓
- X-Request-ID header: `abc123` ✓

## Breaking Changes

None. This fixes behavior that was clearly broken - cookies and headers set before errors should be preserved in error responses.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error responses now consistently apply status and JSON content-type while preserving headers and cookies across error scenarios.
* **Refactor**
  * Internal handling of error/response construction updated for more consistent runtime behavior and stricter type checks.
* **Tests**
  * Added comprehensive tests validating header and cookie preservation, status codes, and messages for various error cases.

### DIFF
--- a/src/dynamic-handle.ts
+++ b/src/dynamic-handle.ts
@@ -361,7 +361,7 @@ export const createDynamicHandler = (app: AnyElysia) => {
 
 					if (response instanceof Promise) response = await response
 
-					// @ts-expect-error jusut in case
+					// @ts-expect-error just in case
 					if (response instanceof ElysiaCustomStatusResponse) {
 						const result = mapEarlyResponse(response, context.set)
 						if (result)


MarcelOlsen:
@coderabbitai review

Issue #1561, opened 2025-11-21
Add support for BunFile type in t.File and t.Files


Teyik0:
### What is the problem this feature would solve?

This would solve the problem in eden treaty when using BunFile having error.
From Elysia perspective it's a feature but from Eden perspective it's a bug.

```ts
const data = await api.categories.post({
	name,
	file: bunfile as File, // current workaround
});
```

Otherwise we get this typescript error ->
```ts
Type 'BunFile' is not assignable to type 'File'.
```

### What is the feature you are proposing to solve the problem?

This is just a small modification on the TFile and TFiles types from Elysia.

PR #1562, opened 2025-11-21
feat: support BunFile on t.Files type system for eden BunFile type error

Teyik0:
Small change to elysia t.File and t.Files allowing eden to infer BunFile properly, close https://github.com/elysiajs/elysia/issues/1561

## Summary by CodeRabbit

* **New Features**
  * Extended file type support to include Bun files alongside standard files, expanding the range of file types the application can process.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,11 +1,12 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "elysia",
       "dependencies": {
         "cookie": "^1.0.2",
-        "exact-mirror": "0.2.2",
+        "exact-mirror": "0.2.3",
         "fast-decode-uri-component": "^1.0.1",
         "memoirist": "^0.4.0",
       },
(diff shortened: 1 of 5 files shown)


Teyik0:
Following this PR, I will also update eden files test (I already did locally will wait this to be merged if ok)
update: Link to the [PR in eden](https://github.com/elysiajs/eden/pull/229) also adding support for auto stringify t.ArrayString and t.ObjectString

SaltyAom:
I would love to add this but unfortunately, this has 2 problems
1. This wouldn't work without `@types/bun`
2. Unwrapping the incoming request with `File | BunFile` would be a breaking change

I would be adding this to Eden side instead, don't worry I can handle that myself

Teyik0:
> I would love to add this but unfortunately, this has 2 problems
> 
> 1. This wouldn't work without `@types/bun`
> 2. Unwrapping the incoming request with `File | BunFile` would be a breaking change
> 
> I would be adding this to Eden side instead, don't worry I can handle that myself

Fair point for node users, thanks in advance !

PR #1568, opened 2025-11-23
fix: Optimize cookie value comparison using FNV-1a hash

YasserGomma:
## Summary

This PR optimizes cookie value comparison by implementing FNV-1a hashing to avoid unnecessary cookie updates when object values haven't changed. This reduces redundant `Set-Cookie` headers and improves performance for applications that frequently set cookie values.

## Problem

Previously, when setting cookie values (especially objects), the implementation would perform expensive `JSON.stringify()` comparisons on every set operation, even when the value hadn't changed. This could lead to:

- Unnecessary `Set-Cookie` headers being sent
- Performance overhead from repeated string serialization
- Potential issues with cookie size limits when dealing with large objects

## Solution

Implemented FNV-1a (Fowler–Noll–Vo) hash-based comparison for object values:

1. **Hash Function**: Added a fast FNV-1a hash implementation that generates a 32-bit hash from the JSON stringified value
2. **Hash Caching**: Store the hash of the current value to enable fast comparison on subsequent sets
3. **Optimized Comparison**: Use hash comparison first, falling back to full JSON comparison only when hashes match (to handle potential collisions)
4. **Performance Optimization**: Cache `JSON.stringify(value)` result to avoid duplicate serialization calls

## Changes

- Added `hashString()` function implementing FNV-1a algorithm
- Added `valueHash` private property to `Cookie` class for caching
- Modified `value` setter to use hash-based comparison for objects
- Optimized to cache JSON.stringify result to avoid duplicate calls
- Improved code comments explaining the hash comparison logic flow and JSON.stringify behavior
- Enhanced test coverage with more realistic scenarios and descriptive test names
- Added test for multiple assignments of the same object in a single request

## Benefits

- **Performance**: Hash comparison is O(1) for comparison (after O(n) hash computation), avoiding expensive string comparisons in most cases
- **Efficiency**: Prevents unnecessary cookie updates when values are unchanged
- **Optimization**: Caches JSON.stringify result to avoid duplicate serialization
- **Scalability**: Better performance with large object values
- **Backward Compatible**: No breaking changes, maintains existing behavior
- **Code Quality**: Improved documentation and test coverage

## Testing

Added comprehensive test cases:
- `should not send set-cookie header when setting same object value as incoming cookie`: Verifies that setting the same object value as an incoming cookie doesn't trigger a Set-Cookie header
- `should not send set-cookie header for large unchanged object values`: Tests performance with large objects (100+ items) to ensure hash comparison works correctly
- `should optimize multiple assignments of same object in single request`: Ensures multiple assignments of the same value in one request only produce one Set-Cookie header

All existing tests continue to pass, ensuring backward compatibility.

## Technical Details

The FNV-1a hash algorithm was chosen for its:
- Fast computation speed
- Good distribution properties
- Simplicity and low memory overhead
- 32-bit output suitable for JavaScript number type

Hash collisions are handled by falling back to full JSON string comparison when hashes match, ensuring correctness while maintaining performance benefits in the common case.

The implementation includes:
- Clear comments explaining when hash comparison is used vs deep comparison
- Note about JSON.stringify key order behavior (objects with same keys in different order are treated as different)
- Proper handling of first-time value sets (when hash is undefined)
- Efficient hash caching to avoid recomputation
- Cached JSON.stringify result to prevent duplicate serialization

## Summary by CodeRabbit

* **Performance**
  * Improved cookie handling to detect unchanged object values and skip redundant set-cookie headers, reducing bandwidth and improving response times for repeated or large cookie updates.
  * Added safe hash caching to optimize repeated assignments while ensuring the cache is invalidated when cookies are updated directly or via update/set.
* **Tests**
  * Added tests verifying unchanged-value skips, large-value optimization, repeated-assignment behavior, and cache invalidation.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -1,4 +1,4 @@
-import type { AnyElysia } from './index'
+import { AnyElysia, Cookie } from './index'
 
 import { Value, TransformDecodeError } from '@sinclair/typebox/value'
 import {
@@ -1573,7 +1573,7 @@ export const composeHandler = ({
 				if (validator.cookie.hasTransform)
 					fnLiteral += coerceTransformDecodeError(
 						`for(const [key,value] of Object.entries(validator.cookie.Decode(cookieValue))){` +
-							`c.cookie[key].value=value` +
+							`c.cookie[key].cookie.value = value` +
 							`}`,
 						'cookie',
 						allowUnsafeValidationDetails
@@ -2131,6 +2131,7 @@ export const composeHandler = ({
 			ERROR_CODE,
 			parseCookie: hasCookie ? parseCookie : undefined,
 			signCookie: hasCookie ? signCookie : undefined,
+			Cookie: hasCookie ? Cookie : undefined,
 			decodeURIComponent: hasQuery ? decode : undefined,
 			ElysiaCustomStatusResponse,
 			ELYSIA_TRACE: hasTrace ? ELYSIA_TRACE : undefined,
--- a/src/cookies.ts
+++ b/src/cookies.ts
@@ -162,7 +162,7 @@ export class Cookie<T> implements ElysiaCookie {
 	}
 
 	protected get setCookie() {
-		if (!(this.name in this.jar)) this.jar[this.name] = { ...this.initial }
+		if (!(this.name in this.jar)) this.jar[this.name] = this.initial
 
 		return this.jar[this.name]
 	}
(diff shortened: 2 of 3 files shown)


akim-bow:
@SaltyAom what do you think about this change?
I personally think that this isn't ideal approach for cookie handling in web frameworks.
https://github.com/elysiajs/elysia/pull/1567 There I left explanation of my POV regarding cookie caching.

PR #1598, opened 2025-12-05
:wrench: fix: use Static<T> for parse/safeParse return types

### LINKED ISSUE #1597: `safeParse()` and `parse()` return schema type instead of static value type

elysia@1.4.17

Darwin 25.1.0 arm64 arm

Bun 1.3.2

no

`safeParse()` returns wrong TypeScript type for data. It returns the schema type instead of the actual value type:

```typescript
import { t } from 'elysia'
import { getSchemaValidator } from 'elysia/schema'

const schema = t.Object({
    id: t.Number(),
    name: t.String()
})

const validator = getSchemaValidator(schema)
const result = validator.safeParse({ id: 1, name: 'test' })

if (result.success) {
    console.log(result.data.id)
    //                 ^^^^
    // Error: Property 'id' does not exist on type 'TObject<...>'
}
```

`result.data` should be typed as `{ id: number, name: string }` (the parsed value)

### What do you see instead?

`result.data` is typed as `TObject<{ id: TNumber, name: TString }>` (the schema definition)

### Additional information

The fix is to use `Static<T>` instead of `T` in return types:

`src/schema.ts`:

```typescript
// Wrong
parse(v: unknown): T
safeParse(v: unknown): { success: true; data: T; ... }

// Correct
parse(v: unknown): Static<T>
safeParse(v: unknown): { success: true; data: Static<T>; ... }
```

`src/types.ts` - Also needs T extends TSchema constraint:

```typescript
// Wrong
export interface ModelValidator<T> extends TypeCheck<T>

// Correct  
export interface ModelValidator<T extends TSchema> extends TypeCheck<T>
```

yes

## Fix implementation

#1598

truehazker:
Fixes #1597 

`parse()` and `safeParse()` now return correct TypeScript types

- Use `Static<T>` instead of `T` for return types in `ElysiaTypeCheck` and `ModelValidator`
- Add `T` extends `TSchema` constraint to ModelValidator
- Fix `safeParse` failure case from `success: true` to `success: false`
- Add `Extract<T, TSchema>` for models getter type constraint
- Add test cases for schema validator type inference

## Summary by CodeRabbit

* **Type System Improvements**
  * Tightened model and validator return types for stricter, more precise validation and parsing results.
  * Public validation APIs now surface a more accurate runtime data shape on successful parse/safeParse.

* **Tests**
  * Added comprehensive unit tests covering parsing, safeParse success/failure, union/discriminated types, and type inference checks.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -994,7 +994,7 @@ export const getSchemaValidator = <
 					error: null
 				}
 			} catch (error) {
-				const errors = [...compiled.Errors(v)].map(mapValueError)
+				const errors = [...validator.Errors(v)].map(mapValueError)
 
 				return {
 					success: false,
@@ -1156,7 +1156,7 @@ export const getSchemaValidator = <
 						error: null
 					}
 				} catch (error) {
-					const errors = [...compiled.Errors(v)].map(mapValueError)
+					const errors = [...validator.Errors(v)].map(mapValueError)
 
 					return {
 						success: false,
@@ -1229,7 +1229,7 @@ export const getSchemaValidator = <
 						error: null
 					}
 				} catch (error) {
-					const errors = [...compiled.Errors(v)].map(mapValueError)
+					const errors = [...validator.Errors(v)].map(mapValueError)
 
 					return {
 						success: false,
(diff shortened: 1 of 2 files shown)


SaltyAom:
I decided to roll my own code because `getSchemaValidator` should also works with Standard Schema as well which #1598 doesn't support

PR #1601, opened 2025-12-05
feat: add built-in logger middleware

### LINKED ISSUE #1599: Add a Built-In Logger Middleware
### What is the problem this feature would solve?

Me as all Elysia users will frequently need basic request/response logging for debugging, profiling, and production visibility.
* every project re-implements its own logger
* the implementations differ slightly (format, timing, status extraction, middleware hooks)
* many users rely on third-party loggers when they only need simple request logs
* newcomers expect logging out of the box like other frameworks (Hono, Express, Fastify)

imo this creates duplication, inconsistent patterns, and makes debugging harder for new users.
a built-in logger would standardize this essential functionality and improve the developer experience.

### What is the feature you are proposing to solve the problem?

what if we introduce a first-class logger middleware built into Elysia, with the following capabilities:
* log HTTP method, path, status code, and response duration
* integrate cleanly with Elysia’s existing lifecycle (derive, onBeforeHandle, onAfterHandle, etc.)
* minimal performance overhead
* zero dependencies
* fully typed with Elysia’s typings

something simple. something that looks like this:
```ts
import { logger } from "elysia/logger"

const app = new Elysia()
  .use(logger())
  .get("/", () => "hello world")

```

### 1. Using third-party logging libraries

These are useful for advanced logging, but overkill for simple request logs.
They also fragment the ecosystem and introduce additional dependencies.

### 2. Manually implementing logging in every project

This leads to copy-pasted logic, inconsistent timing measurements, and slightly different formatting across apps.

### 3. Adding logging only via plugins

Plugins solve part of the problem but don’t address the fact that *logging is a core capability expected from a backend framework* and should be discoverable out of the box.

bossadizenith:
## Summary

Implements a first-class logger middleware for Elysia as proposed in #1599.

### Features
- Log HTTP method, path, status code, and response duration
- Clean integration with Elysia's existing lifecycle hooks (derive, onBeforeHandle, onAfterHandle)
- Minimal performance overhead
- Zero dependencies
- Fully typed with Elysia's typings

### Usage

```typescript
import { logger } from 'elysia/logger'

const app = new Elysia()
  .use(logger())
  .get('/', () => 'hello world')
```

Closes #1599

## Summary by CodeRabbit

* **New Features**
  * Introduced HTTP request logger middleware with customizable output, colored indicators for status codes and HTTP methods, and configurable path skip patterns.

* **Documentation**
  * Added example demonstrating logger integration with a basic server setup.

* **Tests**
  * Added comprehensive test coverage for logger functionality including error handling and path filtering.

### DIFF
--- a/example/logger.ts
+++ b/example/logger.ts
@@ -1,0 +1,17 @@
+import { Elysia } from '../src'
+import { logger } from '../src/logger'
+
+const app = new Elysia()
+	.use(logger())
+	.get('/', () => 'Hello World')
+	.get('/users/:id', ({ params }) => `User ${params.id}`)
+	.post('/users', () => ({ id: 1, name: 'John' }))
+	.get('/health', () => 'OK')
+	.listen(3333)
+
+console.log('🦊 Server running at http://localhost:3333')
+console.log('Try these endpoints:')
+console.log('  GET  http://localhost:3000/')
+console.log('  GET  http://localhost:3000/users/123')
+console.log('  POST http://localhost:3000/users')
+console.log('  GEeT  http://localhost:3000/health')
--- a/package.json
+++ b/package.json
@@ -162,6 +162,11 @@
 			"types": "./dist/universal/file.d.ts",
 			"import": "./dist/universal/file.mjs",
 			"require": "./dist/universal/file.js"
+		},
+		"./logger": {
+			"types": "./dist/logger.d.ts",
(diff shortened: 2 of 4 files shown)


bossadizenith:
it's to mention that more than 50% of the code here was generated with AI however i crossed checked and made sure it doesn't cause harm.

Issue #1138, opened 2025-03-23
Multipart/form-data nested request does not work


Teyik0:
### What is the problem this feature would solve?

All is in the title. We need a way to handle this case out of the box.

 For now we need to do several requests when having nested database model, this has a cost in term of performance and forces us to write more code and unnecessary queries to the database, whereas one unique query would be sufficient with this feature.

Here is a minimal reproduction that of course throw a validation error since Elysia is not handling this case.

```ts
import { treaty } from "@elysiajs/eden";
import { describe, expect, it } from "bun:test";
import { Elysia, t } from "elysia";

const app = new Elysia().post(
  "/",
  ({ body }) => {
    // Anyway what is the logic here, error is happening at validation lifecycle since 
    // Elysia is not parsing the body correctly in this case
    return body;
  },
  {
    body: t.Object({
      name: t.String(),
      // variants is the nested attribute that needs to be parsed correctly by Elysia, without this, it will work.
      variants: t.Array(
        t.Object({
          price: t.Number({ minimum: 0 }),
          weight: t.Number({ minimum: 0 }),
        }),
      ),
      // From the docs ->
      // By providing a file type, Elysia will automatically assume that the content-type is multipart/form-data.
      image: t.File({ type: "image" }),
    }),
  },
);

const api = treaty(app);

const testProduct = {
  name: "Test Product",
  variants: [
    {
      price: 10,
      weight: 100,
    },
  ],
  image: new File(["dummy content"], "image1.webp", { type: "image/webp" }),
};

describe("Product POST/", () => {
  it("should create a product using treaty", async () => {
    const { data, status, error } = await api.index.post(testProduct);
    console.log("Error Value:", error?.value);
    console.log("Error Status:", error?.status);
    expect(status).toBe(200);
    expect(data).toEqual(testProduct);
  });
});
```

### What is the feature you are proposing to solve the problem?

I want Elysia to properly parse multipart/form-data requests when using nested body parameters. This should be the default behavior, as Elysia's type handling does not raise any issues when writing this code boilerplate.

### What alternatives have you considered?

Doing multiple route, with multiple queries to my database.

Teyik0:
Thank to @SaltyAom help, found a workaround using t.ArrayString(), updated sample code below for people that had this problem.

Treaty will complain for a ts-error without the unknown but it work 👍 

```typescript
import { treaty } from "@elysiajs/eden";
import { describe, expect, it } from "bun:test";
import { Elysia, t } from "elysia";

[...]

PR #1607, opened 2025-12-07
Fix nested formdata

### LINKED ISSUE #1606: t.ObjectString default value does not work and t.Optional doest not work with t.ObjectString

1.4.18

Darwin 24.6.0 arm64 arm

1.3.4

no

All this is somehow related to #1138 

First default value for ObjectString ->

```ts
it("Default value", async () => {
	const schema = t.ObjectString(
		{
			pageIndex: t.Number(),
			pageLimit: t.Number(),
		},
		{
			default: { pageIndex: 0, pageLimit: 10 },
		},
	);

	expect(Value.Create(schema)).toEqual({ pageIndex: 0, pageLimit: 10 }); // does not pass, default is '{}'
});
```

Second t.Optional with t.ObjectString ->

```ts
const schema = t.Object({
	name: t.String(),
	metadata: t.Optional(
		t.ObjectString({
			pageIndex: t.Number(),
			pageLimit: t.Number(),
		}),
	),
});

// Here all working as expected
it("Optional", async () => {
	expect(Value.Check(schema, { name: "test" })).toBe(true);
	expect(Value.Create(schema).metadata).toBeUndefined();

	expect(
		Value.Check(schema, {
			name: "test",
			metadata: { pageIndex: 1, pageLimit: 10 },
		}),
	).toBe(true);
	expect(Value.Check(schema, { name: "test", metadata: {} })).toBe(false);
});

// Here it is failing when somehow send to Elysia
it("Fail when using elysia", async () => {
	const app = new Elysia().post("/", ({ body }) => body, {
		body: schema,
	});
	const response = await app.handle(
		new Request("http://localhost/", {
			method: "POST",
			headers: {
				"Content-Type": "application/json",
			},
			body: JSON.stringify({
				name: "test",
			}),
		}),
	);
	const result = await response.json();
	console.log(result);

	expect(response.status).toBe(200); // return a 422 validation
/*
Here is the validation error ->
{
  type: "validation",
  on: "property",
  property: "/pageIndex",
  message: "Expected required property",
  summary: "Property 'pageIndex' is missing",
  found: {},
  errors: [
    {
      type: 45,
      schema: [Object ...],
      path: "/pageIndex",
      message: "Expected required property",
      errors: [],
      summary: "Property 'pageIndex' is missing",
    }, {
      type: 45,
      schema: [Object ...],
      path: "/pageLimit",
      message: "Expected required property",
      errors: [],
      summary: "Property 'pageLimit' is missing",
    }, {
      summary: "Expected number",
      type: 41,
      schema: [Object ...],
      path: "/pageIndex",
      message: "Expected number",
      errors: [],
    }, {
      summary: "Expected number",
      type: 41,
      schema: [Object ...],
      path: "/pageLimit",
      message: "Expected number",
      errors: [],
    }
  ],
}
41 | 		}),
42 | 	);
43 | 	const result = await response.json();
44 | 	console.log(result);
45 |
46 | 	expect(response.status).toBe(200);
                              ^
error: expect(received).toBe(expected)

Expected: 200
Received: 422
*/
});

// Obviously work when all options are set
it("Work when using elysia", async () => {
	const app = new Elysia().post("/", ({ body }) => body, {
		body: schema,
	});
	const response = await app.handle(
		new Request("http://localhost/", {
			method: "POST",
			headers: {
				"Content-Type": "application/json",
			},
			body: JSON.stringify({
				name: "test",
				metadata: { pageIndex: 1, pageLimit: 10 },
			}),
		}),
	);
	const result = await response.json();
	console.log(result);

});
```

Elysia should work when using t.ObjectString paired with t.Optional

### What do you see instead?

_No response_

### Additional information

The problem comes in the Elysia internal not the ObjectString itself. This may happen in replaceSchema internal method.
root cause -> _replaceSchemaType -> composeProperties -> _composeProperties

_No response_

Teyik0:
This fix #1606 and #1138 by adding coerceFormdata capabilities.

At first I just wanted to fix #1606 when I saw that t.Optional was'nt working with t.ObjectString but when changing t.ObjectString default value from '{}' to options.default it was breaking t.ArrayQuery. I was in a dead end.

Then after looking at the code, I understood that Elysia coercion is amazing and that I just needed that out of the box for nested formdata. Thus, I wanted to fix replaceSchema function and somehow end up rewriting it to make it more readable and functional. The root cause was somewhere in the composeProperties function, but in any case with the rewrite it's not needed anymore.

Finally, since the rewrite was working and fixing 1606, I just needed to add a coerceFormdata function such as stringToStructureCoercions for ArrayQuery. And of course a bunch of new tests (most of the PR is new test).

The only drawback is that coerceFormdata is only working for elysia type system. But all already coercion is only working for elysia type system, thus I think it's ok like this.

I hope you will take some time to review it, and, in any case I learned so much reading the code base and for that big thanks!

## Summary by CodeRabbit

* **New Features**
  * Enhanced validation for form data submissions with file uploads and complex nested structures.

* **Bug Fixes**
  * Improved handling of union-based schemas during body validation, ensuring properties are correctly extracted and processed.

* **Tests**
  * Added comprehensive test coverage for FormData handling with file uploads and nested object validation.

### DIFF
--- a/test/type-system/formdata.test.ts
+++ b/test/type-system/formdata.test.ts
@@ -3,15 +3,15 @@ import { z } from "zod";
 import { Elysia, fileType, t, type ValidationError } from "../../src";
 
 const variantObject = t.Object({
-			price: t.Number({ minimum: 0 }),
-			weight: t.Number({ minimum: 0 }),
-		})
+	price: t.Number({ minimum: 0 }),
+	weight: t.Number({ minimum: 0 }),
+});
 
 const metadataObject = {
-		category: t.String(),
-		tags: t.Array(t.String()),
-		inStock: t.Boolean(),
-	}
+	category: t.String(),
+	tags: t.Array(t.String()),
+	inStock: t.Boolean(),
+};
 
 const postProductModel = t.Object({
 	name: t.String(),
@@ -23,9 +23,8 @@ type postProductModel = typeof postProductModel.static;
 
 const patchProductModel = t.Object({
 	name: t.Optional(t.String()),
-	variants: t.Optional(t.Array(variantObject	),
-	),
-	metadata: t.Optional(t.Object(metadataObject),),
+	variants: t.Optional(t.Array(variantObject)),
+	metadata: t.Optional(t.Object(metadataObject)),
 	image: t.Optional(t.File({ type: "image" })),
 });
 type patchProductModel = typeof patchProductModel.static;
@@ -40,14 +39,16 @@ type postProductModelComplex = typeof postProductModelComplex.static;
 
 const patchProductModelComplex = t.Object({
 	name: t.Optional(t.String()),
-	variants: t.Optional(t.ArrayString(variantObject	),
-	),
-	metadata: t.Optional(t.ObjectString(metadataObject),),
+	variants: t.Optional(t.ArrayString(variantObject)),
+	metadata: t.Optional(t.ObjectString(metadataObject)),
 	image: t.Optional(t.File({ type: "image" })),
 });
 type patchProductModelComplex = typeof patchProductModelComplex.static;
 
-describe.each([{ aot: true }, { aot: false }])("Nested FormData with file(s) support (aot: $aot)", ({ aot }) => {
+describe.each([
+	{ aot: true },
+	{ aot: false },
+])("Nested FormData with file(s) support (aot: $aot)", ({ aot }) => {
 	const app = new Elysia({ aot })
 		.post("/product", async ({ body, status }) => status("Created", body), {
 			body: postProductModel,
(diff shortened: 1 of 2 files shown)


src/schema.ts
```diff
@@ -122,36 +123,63 @@ export const hasAdditionalProperties = (
 	return false
 }
 
-export const hasType = (type: string, schema: TAnySchema) => {
+/**
+ * Resolve a schema that might be a model reference (string) to the actual schema
+ */
+ export const resolveSchema = (
```
Teyik0, line 129: This is needed in order to resolve string imported model. Otherwise, t.model for example would not use coerceFormdata when needed.

src/schema.ts
```diff
@@ -122,36 +123,63 @@ export const hasAdditionalProperties = (
...
+	// @ts-expect-error private property
+	if (modules && schema in modules.$defs) {
+		return (modules as TModule<{}, {}>).Import(schema as never)
+	}
+
+	// Then check models
+	return models?.[schema]
+ }
+
+export const hasType = (type: string, schema: TAnySchema): boolean => {
```
Teyik0, line 147: I somehow enhanced all edgecase that was previsouly not handled on this function

PR #1624, opened 2025-12-14
fix: apply macros before merging hooks in group method

### LINKED ISSUE #1586: `macro.beforeHandle` runs after nested plugin `resolve` hooks despite same-queue documentation

1.4.16

Darwin 25.1.0 arm64 arm

1.3.1

No

```js

// Step 1: Auth service with resolve + macro
const authService = new Elysia({ name: 'auth-service' })
  .resolve(() => {
    console.log('1. authService.resolve')
    return { userId: undefined } // Simulating no auth
  })
  .macro({
    isSignedIn: {
      beforeHandle({ userId }) {
        console.log('3. isSignedIn.beforeHandle')
        if (!userId) throw new Error('Unauthorized')
      },
    },
  })
  .as('scoped')

// Step 2: DB client that requires userId
const dbClient = new Elysia({ name: 'db-client' })
  .resolve((ctx) => {
    console.log('2. dbClient.resolve')
    const userId = (ctx as { userId?: string }).userId
    if (!userId) throw new Error('User ID is required') // ❌ Throws here
    return { db: { userId } }
  })
  .as('scoped')

// Step 3: Feature module using dbClient
const feature = new Elysia({ name: 'feature' })
  .use(dbClient)
  .get('/', ({ db }) => `Hello ${db.userId}`)

// Step 4: Main app
const app = new Elysia()
  .use(authService)
  .group('/v1', { isSignedIn: true }, (app) => app.use(feature))
  .listen(3000)
```

With the example above:

1/ curl http://localhost:3000/v1/
2/ observe error "User ID is required", expected would be "Unauthorized"

When a group has a macro that defines a `beforeHandle` hook, it should run before any `resolve` hooks from plugins used inside that group, since the guard is registered at the group level (before nested plugins).

### What do you see instead?

The nested plugin's `resolve` runs before the macro's `beforeHandle`, causing errors before the guard can reject requests.

### Additional information

According to the [documentation](https://elysiajs.com/essential/life-cycle#queue:~:text=Using%20resolve%20and%20onBeforeHandle%20is%20stored%20in%20the%20same%20queue), `beforeHandle` and `resolve` belong to the same execution queue and their order depends on plugin registration order. However, when using a macro-defined `beforeHandle` within a group, nested plugins' `resolve` hooks execute before the macro's `beforeHandle`.

Yes

tt-a1i:
## Summary

Fixes #1586 - macro-defined `beforeHandle` hooks were running after nested plugin `resolve` hooks instead of before.

**Root Cause**: The `group` method was merging hooks via `mergeHook(hook, localHook)` before expanding macro options. The macro expansion happened later in `add()`, but by then the nested plugin hooks were already in the array, so macro hooks were appended at the end.

**Fix**: 
1. Call `applyMacro(hook)` before `mergeHook()` in the group method (similar to how `guard` method already does this at line 4504)
2. Properly merge `hook.standaloneValidator` (from macro schema) with `localHook.standaloneValidator` to prevent macro-defined schema validation from being lost

## Test Cases Added

Four test cases in `test/core/macro-lifecycle.test.ts`:

1. **macro beforeHandle runs before nested plugin resolve**: Verifies that when a group uses `{ isSignedIn: true }` macro, the macro's `beforeHandle` runs before any nested plugin's `resolve` hook
2. **hooks run in registration order**: Verifies that `resolve` and `beforeHandle` share the same queue and run in registration order
3. **macro in group with successful auth**: Verifies the correct execution order when auth succeeds
4. **preserve macro schema when merging**: Verifies that macro-defined schema validation is preserved when used with nested plugins

## Before Fix
```
Execution order: [ "auth.resolve", "data.resolve", "requireAuth.beforeHandle" ]
```

## After Fix
```
Execution order: [ "auth.resolve", "requireAuth.beforeHandle", "data.resolve" ]
```

## Test Results
- All existing tests pass (except pre-existing flaky `stop.test.ts`)
- All 4 new tests pass
- All macro and lifecycle related tests verified

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured macros are expanded before group merging, fixing execution order in grouped routes and improving authorization and dependency resolution.
  * Corrected merging of standalone validators so combined validators from macros, local hooks, and group schemas are applied consistently.

* **Tests**
  * Added comprehensive tests validating macro and hook lifecycle ordering across grouped routes, registration order, and authorization flows.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4066,20 +4066,29 @@ export default class Elysia<
 											localHook.error,
 											...(sandbox.event.error ?? [])
 										],
-							standaloneValidator: !hasStandaloneSchema
-								? localHook.standaloneValidator
-								: [
-										...(localHook.standaloneValidator ??
-											[]),
-										{
-											body,
-											headers,
-											query,
-											params,
-											cookie,
-											response
-										}
-									]
+							// Merge macro's standaloneValidator with local and group schema
+							standaloneValidator:
+								hook.standaloneValidator ||
+								localHook.standaloneValidator ||
+								hasStandaloneSchema
+									? [
+											...(hook.standaloneValidator ?? []),
+											...(localHook.standaloneValidator ??
+												[]),
+											...(hasStandaloneSchema
+												? [
+														{
+															body,
+															headers,
+															query,
+															params,
+															cookie,
+															response
+														}
+													]
+												: [])
+										]
+									: undefined
 						}),
 						undefined
 					)
--- a/test/core/macro-lifecycle.test.ts
+++ b/test/core/macro-lifecycle.test.ts
@@ -139,4 +139,54 @@ describe('macro beforeHandle lifecycle order', () => {
 			'data.resolve'
 		])
 	})
+
+	it('should preserve macro schema when merging with nested plugin hooks', async () => {
+		const { t } = await import('../../src')
+
+		// Macro that adds both beforeHandle and body schema
+		const validatedMacro = new Elysia({ name: 'validated-macro' })
+			.macro({
+				validatePayload: {
+					body: t.Object({
+						name: t.String()
+					}),
+					beforeHandle() {
+						// Macro's beforeHandle
+					}
+				}
+			})
+			.as('scoped')
+
+		const nestedPlugin = new Elysia({ name: 'nested' })
+			.resolve(() => ({ nested: true }))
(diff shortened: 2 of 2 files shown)


heiwen:
@SaltyAom do you think this is the right fix?

PR #1634, opened 2025-12-23
:wrench: fix: t.Record validation failure in response schemas

### LINKED ISSUE #1631: Request validation errors with specific schema structure

elysia@1.4.19

Linux 6.12.59 x86_64 unknown / elysia playground

1.3.2

no

have a file like this:

```ts
import { Elysia, t } from "elysia"

export const app = new Elysia()
	.get(
		"",
		() => {
			return {
				list: [
					{
						toto: {
							bar: 1,
						},
						foo: {
							link: "first",
						},
					},
				],
				one: {
					toto: {
						bar: 0,
					},
					foo: {
						link: "second",
					},
				},
			};
		},
		{
			response: {
				200: t.Object({
					list: t.Array(
						t.Object({
							toto: t.Object({
								bar: t.Integer(),
							}),
							foo: t.Record(t.String(), t.String()),
						}),
					),
					one: t.Object({
						toto: t.Object({
							bar: t.Integer(),
						}),
						foo: t.Record(t.String(), t.String()),
					}),
				})
			},
		},
	)
    .listen(3000);
```

and request it `curl http://localhost:3000`

to retrieve this:
```json
{
    "list": [
        {
            "toto": {
                "bar": 1
            },
            "foo": {
                "link": "first"
            }
        }
    ],
    "one": {
        "toto": {
            "bar": 0
        },
        "foo": {
            "link": "second"
        }
    }
}
```

### What do you see instead?

```json
{
  "type": "validation",
  "on": "response",
  "property": "root",
  "expected": {
    "list": [],
    "one": {
      "toto": {
        "bar": 0
      },
      "foo": {}
    }
  },
  "found": {
    "list": [
      {
        "toto": {
          "bar": 1
        },
        "foo": {
          "link": "first"
        }
      }
    ],
    "one": {
      "toto": {
        "bar": 0
      },
      "foo": {
        "link": "second"
      }
    }
  },
  "errors": []
}
```

### Additional information

If the JSON schema isn't exactly like the one i sent this works normally. You can't remove a property, the array nor the nested objects. You can add properties and the error will stay the same tho.

I got to this schema by removing properties from my models one by one to see what caused the issue and i have no idea what might be the issue ngl.

yes, also tried on the playground

MarcelOlsen:
# Fix t.Record validation failure in response schemas

this resolves #1631 

## Problem

Response validation fails when using `t.Record` types, returning 422 validation errors even with valid data.

**Example:**
```typescript
const app = new Elysia().get('/', () => ({
    list: [{
        toto: { bar: 1 },
        foo: { link: 'first' }
    }],
    one: {
        toto: { bar: 0 },
        foo: { link: 'second' }
    }
}), {
    response: {
        200: t.Object({
            list: t.Array(t.Object({
                toto: t.Object({ bar: t.Integer() }),
                foo: t.Record(t.String(), t.String())
            })),
            one: t.Object({
                toto: t.Object({ bar: t.Integer() }),
                foo: t.Record(t.String(), t.String())
            })
        })
    }
})

// Returns 422 with validation error
// Should return 200 with the data
```

## Root Cause

The bug is in `exact-mirror` library - `optionalsInArray` cleanup state was bleeding across sibling structures, causing Record properties to be incorrectly removed.

### Upstream Fix (exact-mirror)
Fixed in elysiajs/exact-mirror#26 by clearing `optionalsInArray` state after use, preventing pollution across sibling arrays/records.

### Compatibility Changes (this PR)

**src/schema.ts:**
- Updated `createMirror` calls to use `TAnySchema` type casts instead of `TSchema` (lines 543, 805, 944)
- Ensures compatibility with exact-mirror's stricter type signature

**test/validator/response.test.ts:**
- Added regression test for Record with nested objects (lines 566-593)
- Ensures the original bug scenario stays fixed

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -540,7 +540,7 @@ export const getSchemaValidator = <
 			let mirror: Function
 			if (normalize === true || normalize === 'exactMirror')
 				try {
-					mirror = createMirror(schema as TSchema, {
+					mirror = createMirror(schema as TAnySchema, {
 						TypeCompiler,
 						sanitize: sanitize?.(),
 						modules
@@ -802,7 +802,7 @@ export const getSchemaValidator = <
 			if (normalize && !hasPatternProperties(schema)) {
 				if (normalize === true || normalize === 'exactMirror') {
 					try {
-						validator.Clean = createMirror(schema, {
+						validator.Clean = createMirror(schema as TAnySchema, {
 							TypeCompiler,
 							sanitize: sanitize?.(),
 							modules
@@ -941,7 +941,7 @@ export const getSchemaValidator = <
 		if (normalize && !hasPatternProperties(schema)) {
 			if (normalize === true || normalize === 'exactMirror') {
 				try {
-					compiled.Clean = createMirror(schema, {
+					compiled.Clean = createMirror(schema as TAnySchema, {
 						TypeCompiler,
 						sanitize: sanitize?.(),
 						modules
--- a/test/validator/response.test.ts
+++ b/test/validator/response.test.ts
@@ -563,145 +563,32 @@ describe('Response Validator', () => {
 		expect(result.join('')).toContain('data: {"name":"Name"}')
 	})
 
-	it('validate Record with nested objects (original bug)', async () => {
-		const app = new Elysia().get(
-			'',
-			() => ({
-				list: [
-					{
-						toto: { bar: 1 },
-						foo: { link: 'first' }
-					}
-				],
-				one: {
-					toto: { bar: 0 },
-					foo: { link: 'second' }
-				}
-			}),
-			{
-				response: {
-					200: t.Object({
-						list: t.Array(
-							t.Object({
-								toto: t.Object({ bar: t.Integer() }),
-								foo: t.Record(t.String(), t.String())
-							})
-						),
-						one: t.Object({
-							toto: t.Object({ bar: t.Integer() }),
-							foo: t.Record(t.String(), t.String())
-						})
-					})
-				}
-			}
-		)
-
-		const res = await app.handle(req('/'))
-		const json = await res.json()
-
-		// Should return 200, not 422 validation error
-		expect(res.status).toBe(200)
-		expect(json).toEqual({
-			list: [
-				{
-					toto: { bar: 1 },
-					foo: { link: 'first' }
-				}
-			],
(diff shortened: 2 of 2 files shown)


MarcelOlsen:
> **Actionable comments posted: 0**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > src/schema.ts (1)> `130-130`: **Fix: `schema.not` is a single schema, not an array.**
> > > Line 130 incorrectly treats `schema.not` as an array by calling `.some()` on it, which will cause a runtime error. According to JSON Schema specification and TypeBox, `not` accepts a single schema object, not an array.
> > > This is the same bug that was fixed in `hasPatternProperties` at line 100. The correct pattern (used at line 329 in `hasRef` and line 377 in `hasTransform`) is to call the predicate function directly.
> > > 
> > > 🔎 Proposed fix
> > > ```diff
> > > -	if (schema.not) return schema.not.some(hasAdditionalProperties)
> > > +	if (schema.not) return hasAdditionalProperties(schema.not)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > Based on past review comments and the fix applied to `hasPatternProperties`.
> 
> 📜 Review details

This is marked outside of the scope and the change has not been introduced in this PR, so this will not be getting addressed

zoriya:
just for information, i have routes that use t.Record without this issue so I'm not sure if this is really the issue.

i also don't think that disabling object cleaning when there's a record anywhere in the object is a correct fix. it should clean the object, just not the record.

MarcelOlsen:
Hmm, yeah, now that I think about it this does seem kinda too broad. Also, this might actually be a similar issue to what LASkuma was facing after all. Would you be able to check if this draft PR https://github.com/elysiajs/exact-mirror/pull/26 fixes your issue? It seems to return the correct thing on my end @zoriya:
```bash
> curl http://localhost:3003/

{"list":[{"toto":{"bar":1},"foo":{"link":"first"}}],"one":{"toto":{"bar":0},"foo":{"link":"second"}}}%  
```

Thanks for the feedback, nice catch!

SaltyAom:
my bad, I accidentally click merge but I think it should not

PR #1643, opened 2025-12-29
fix: restore handle() return type to Promise<Response>

### LINKED ISSUE #1630: handle used to return a Promise<Response> not it returns Promise<any>

1.4.19

MacOS

1.3.2

no

Write this code 
```
const response = await app.handle(
      new Request("http://localhost/test"),
    );

```

In pre 1.4.19 response was typed as `Response` object, no is any, which means that when one upgrade to 1.4.19, I get a bunch of type errors 
```
error  Unsafe member access .status on an `any` value  @typescript-eslint/no-unsafe-member-access
``` 

Key the same type or if its a breaking change then use another version, not patch.

### What do you see instead?

Saw a breaking change it should keep `Promise<Response>` type for `handle` return type.

### Additional information

_No response_

yes

eL1fe:
## Summary
- Fixes #1630
- Restores the `handle()` method return type from `Promise<any>` to `Promise<Response>`
- Also adds explicit return type to the `fetch` getter

## Problem
The `handle` method was returning `Promise<any>` instead of `Promise<Response>` because:
1. `composeGeneralHandler` creates a function via `Function()` constructor, which returns `any`
2. Neither `handle` nor `fetch` had explicit return type annotations
3. TypeScript inferred the type as `any`

## Solution
Added explicit return types:
- `handle` now returns `Promise<Response>`
- `fetch` getter now returns `(request: Request) => Response | Promise<Response>`

## Test plan
- [x] Build passes
- [x] Existing tests pass (23 pre-existing failures unrelated to this change)
- [ ] Type inference verified with reproduction from #1630

## Summary by CodeRabbit

* **Refactor**
  * Enhanced API type signatures with explicit return type annotations to improve type safety and provide clearer contracts for request handling and response processing.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8044,14 +8044,14 @@ export default class Elysia<
 		return this
 	}
 
-	handle = async (request: Request) => this.fetch(request)
+	handle = async (request: Request): Promise<Response> => this.fetch(request)
 
 	/**
 	 * Use handle can be either sync or async to save performance.
 	 *
 	 * Beside benchmark purpose, please use 'handle' instead.
 	 */
-	get fetch() {
+	get fetch(): (request: Request) => Response | Promise<Response> {
 		const fetch = this.config.aot
 			? composeGeneralHandler(this)
 			: createDynamicHandler(this)


SaltyAom:
this was fixed somewhere between #1630, I forgot which one but it doesn't happend anymore, so closing as complete

PR #1647, opened 2025-12-30
fix: resolve "Body already used" when accessing request in derive/resolve hooks

### LINKED ISSUE #1628: Body already used error in v1.4.19 when accessing request.body or request.arrayBuffer()

elysia@1.4.19

Bun version: 1.3.5

No, I am not using new Elysia({ aot: false }).

**Steps:**

- Create a new Elysia app with version 1.4.19
- Add a plugin using .derive() that attempts to read request.arrayBuffer() or access request.body
- Make a POST request with a JSON body
- Observe the error

```

const rawBodyPlugin = new Elysia({ name: 'raw-body' })
  .derive(async ({ request }) => {
    let rawBody: ArrayBuffer | undefined;

    if (request.method !== 'GET' && request.method !== 'HEAD') {
      try {
        rawBody = await request.arrayBuffer();
      } catch (err) {
        console.error('Failed to capture body:', err);
        rawBody = undefined;
      }
    }
    return { rawBody };
  })
  .as('scoped');

const app = new Elysia()
  .use(rawBodyPlugin) // First plugin in the chain
  .post('/test', ({ rawBody }) => {
    return { bodySize: rawBody?.byteLength ?? 0 };
  })
  .listen(3000);
```

In version 1.4.18 and earlier, it's possible to be able to access request.arrayBuffer() or request.body without issues, since nothing has consumed the body stream yet.

The body should be readable by the first plugin that attempts to access it.

### What do you see instead?

```
TypeError: Body already used
 code: "ERR_BODY_ALREADY_USED"
      at <anonymous> (src/plugins/raw-body.plugin.ts:15:33)
      at <anonymous> (src/plugins/raw-body.plugin.ts:9:11)
      at handle (node_modules/elysia/dist/bun/index.js:34:33)
```

### Additional information

- Workaround: Downgrade to elysia@1.4.18 - the bug does not occur in this version.
- This is critical for API Gateway use cases where the body must be forwarded to upstream services.
- Using request.clone() before reading also fails with the same error, indicating the body is consumed at the framework level before user code runs.
- Setting parse: [] on routes to disable body parsing does not prevent the issue.

Yes, I ran rm -rf node_modules && rm bun.lockb && bun install and the issue persists with v1.4.19. Downgrading to v1.4.18 is the only working solution.

eL1fe:
## Summary
- Fixes #1628 - "Body already used" error when calling `request.arrayBuffer()` in derive hooks
- Adds `request` tracking to Sucrose inference to detect when hooks access the request body
- Clones request before parsing so derive/resolve hooks can still read from it
- Adds missing derive and resolve hooks to sucrose analysis

## Problem
When a derive() hook tries to access `request.arrayBuffer()`, `request.text()`, or `request.json()`, it fails with "Body already used" because Elysia parses the body BEFORE running derive hooks:

```typescript
.derive(async ({ request }) => {
  const rawBody = await request.arrayBuffer() // Error: Body already used
  return { rawBody }
})
```

## Solution
1. Detect when hooks access `request` via Sucrose inference
2. When detected, clone the request before body parsing
3. Parse body from the cloned request's rawBody
4. Keep original request available for derive/resolve hooks

## Test plan
- [x] Verified fix resolves the issue from #1628
- [x] All existing tests pass (1403 pass, 23 pre-existing failures unrelated to this change)
- [x] Verified inline parse hooks still work correctly

## Summary by CodeRabbit

* **New Features**
  * Preserve the raw request body when inference-based request handling is enabled and prefer it for decoding JSON, text, urlencoded and binary payloads; multipart/form-data remains handled specially.
  * Ensure custom parsers receive normalized content-type and surface parsing errors consistently.

* **Chores**
  * Added a new inference flag to track explicit request access.

* **Tests**
  * Updated tests to include the new request-inference flag.

### DIFF
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -919,7 +919,10 @@ export const composeHandler = ({
 		// Skip if custom parse hooks exist - they'll read the request themselves
 		// Note: FormData requests cannot be cloned for reuse - derive/resolve cannot
 		// re-read the body for multipart/form-data requests (protocol limitation)
-		if (inference.request && !hooks.parse?.length) {
+		const hasCustomParseHook = hooks.parse?.some(
+			(h: any) => typeof h !== 'string' && typeof h.fn !== 'string'
+		)
+		if (inference.request && !hasCustomParseHook) {
 			fnLiteral +=
 				`const _ct=c.request.headers.get('content-type')\n` +
 				`if(!_ct||!_ct.includes('multipart/form-data')){` +
--- a/src/compose.ts
+++ b/src/compose.ts
@@ -914,6 +914,18 @@ export const composeHandler = ({
 
 		fnLiteral += '\ntry{'
 
+		// When hooks access request.body/arrayBuffer, preserve raw body before parsing
+		// Clone request first so hooks can still read from it
+		// Skip if custom parse hooks exist - they'll read the request themselves
+		// Note: FormData requests cannot be cloned for reuse - derive/resolve cannot
+		// re-read the body for multipart/form-data requests (protocol limitation)
+		if (inference.request && !hooks.parse?.length) {
+			fnLiteral +=
+				`const _ct=c.request.headers.get('content-type')\n` +
+				`if(!_ct||!_ct.includes('multipart/form-data')){` +
+				`c.rawBody=await c.request.clone().arrayBuffer()}\n`
+		}
+
 		let parser: string | undefined =
 			typeof hooks.parse === 'string'
 				? hooks.parse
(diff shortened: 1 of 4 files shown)


joebnb:
@SaltyAom would review and merge this? i also found this problem,and rise a issue to logger, today when i adding zod schema found even logger not enable, it's also happen,maybe this function should not use promise to implement

https://github.com/bogeychan/elysia-logger/issues/43

SaltyAom:
@joebnb, please see https://github.com/elysiajs/elysia/issues/1628#issuecomment-3707076900

alanbucknum:
@SaltyAom is this proposed solution feasible? It would help our implementation of Elysia, we have been unable to upgrade our elysia version because of this problem.

SaltyAom:
@alanbucknum can you share an example code that can replicate the problem?

pjrebsch:
@SaltyAom Replying on behalf of @alanbucknum, I was able to reproduce.  Though, this may technically be a different issue leading to the same error.  I posted the reproduction details here: https://github.com/elysiajs/elysia/issues/1628#issuecomment-4100351707

eL1fe:
@pjrebsch your repro case (passing ctx to a helper that reads request body) is covered now too.

In AOT sucrose's isContextPassToFunction now sets inference.request when the whole context goes to a helper, so request gets cloned before body parsing. In dynamic mode the handler clones the request before parsing so the original stays available.

Tested with your exact code, both modes work.

eL1fe:
Rebased this on latest main and resolved the body parsing conflicts

PR #1651, opened 2026-01-03
types: Mark app.server as non-null in START event handler

kiriDevs:
## What?

This PR changes the typing of the handlers for the `start` event, declaring `app.server` as non-null for them.

It also updates the example snipped on `onStart`, removing the optional-chaining that is not needed because of this change anymore.

## Why?

A miniscule, personal, stylistic annoyance:

```typescript
new Elysia().onStart((startedApp) => {
  console.log("Listening on", startedApp.server?.url.host);
});
```

> This is the handler for after the server has started, why do I still have to check the server exists?

After this change, this will work just fine:
```typescript
console.log("Listening on", startedApp.server.url.host);
```
(The optional-chaining ("`null`-check") is no longer needed)

I guess another option would be using `app.listen`'s callback, but that's only feasible for `onStart` event handlers on the "main" `Elysia` instance (i.e. outside of plugins).

---

As far as I was able to tell the `"start"` event handlers are never called from outside of `listen`, where `app.server` is set with `Bun.serve(…)` just before. Since `Bun.serve(…)` does not return `null` (or `undefined`) (according to TypeScript, and the docs don't mention anything contrary), this should be a safe guarantee to make. For what it's worth, `tsc` (`bun run test:types`) doesn't have any issues with the changes either.

I completely understand if this can't/won't be merged, be it because it's intentionally typed the current way for future changes/expansions or just because it's super insignificant; as mentioned, it was just a super small annoyance to me and I figured it'd be easy (and quick) enough to change.

## Summary by CodeRabbit

* **New Features**
  * Added a new start lifecycle handler type and updated the start hook signature.
  * Start handlers now always receive a non-optional server object (server.url, server.port available).

* **Documentation**
  * Updated examples and JSDoc to reflect the guaranteed server property in start handlers.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -8275,7 +8275,8 @@ export type {
 	TransformHandler,
 	HTTPHeaders,
 	EmptyRouteSchema,
-	ExtractErrorFromHandle
+	ExtractErrorFromHandle,
+        ServerStartHandler
 } from './types'
 
 export { env } from './universal/env'


SaltyAom:
Technically it could still be nullable if you doesn't run Elysia server in Bun

kiriDevs:
Fair enough, I hadn't considered adapters outside of this repo (and saw that, from the four inside this repo, only the `bun` one implements `listen()` further than `throw new Error`). Sorry; I guess this isn't feasible (or even desired) then?

PR #1719, opened 2026-02-02
Add t.Union/t.Intersection handling in property enumerations/checks

coreh:
Hit this error:

```
TypeError: undefined is not an object (evaluating 'Object.keys(properties)')
    at keys (unknown)\n    at <anonymous> (/app/node_modules/elysia/src/dynamic-handle.ts:416:36)
    at processTicksAndRejections (native:7:39)
```

Looking into it, realized elysia is doing naive `.properties`-based checks and iteration in a couple of places, which meant that routes using `t.Union`/`t.Intersect` would either crash (my case) or silently behave incorrectly.

Also the files code was handling `.anyOf` but not `.allOf` or `.oneOf`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Safer, more robust handling of composite schemas (unions/intersections) across query, body, cookie and file validation.
  * Prevented crashes and improved default/value detection when schema properties are absent or nested.
  * Improved reliability of file upload handling in complex schema compositions.

* **Tests**
  * Added extensive tests for unions, intersections, file uploads, and query coercion/array scenarios.

### DIFF
--- a/src/schema.ts
+++ b/src/schema.ts
@@ -1077,8 +1077,8 @@ export const getSchemaProperties = (
 
 	if (schema.properties) return schema.properties
 
-	if (schema.allOf || schema.anyOf) {
-		const members = schema.allOf ?? schema.anyOf
+	const members = schema.allOf ?? schema.anyOf ?? schema.oneOf
+	if (members) {
 		const result: Record<string, TAnySchema> = {}
 		for (const member of members) {
 			const props = getSchemaProperties(member)


src/schema.ts
```diff
@@ -745,7 +758,7 @@ export const getSchemaValidator = <
 						hasAdditionalProperties(schema))
 				},
 				get hasDefault() {
-					if ('~hasDefault' in this) return this['~hasDefault']
```
coreh, line 748: Drive-by fix, not sure if it's something in my setup but typescript was complaining about this, the other methods had `!` so I added it in

PR #1763, opened 2026-02-27
fix: Refactor ValueToResponseSchema type logic to account for return type `something | undefined`

### LINKED ISSUE #1762: With a response being `something | undefined` elysia/eden always returns unknown as data type

1.4.26

Linux 6.19.2-200.nobara.fc43.x86_64 x86_64 unknown

bun 1.3.9

No

```ts
import { treaty } from "@elysiajs/eden/treaty2";
import Elysia from "elysia";

const app = new Elysia().get("/", () => "Hello World" as string | undefined);

type App = typeof app;

const rsp = await treaty<App>("").get();

rsp.data;
//   ^? (property) data: unknown
```

`rsp.data` should have type `string | undefined`

### What do you see instead?

`rsp.data` is typed as `unknown`

### Additional information

This happens because of `ValueToResponseSchema` in `src/types.ts` which does not infer the 200 schema from the return type if it extends undefined.
I will propose a fix for this in a PR.

Yes

deadlinecode:
Refactor ValueToResponseSchema type to ensure only returning empty 200 schema on `undefined` or `void`. If it is `something | undefined` type response as infered return type.

Should fix #1762 

## Summary by CodeRabbit

* **Refactor**
  * Improved internal type handling logic for response schema generation to ensure more precise type definitions.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2185,7 +2185,7 @@ export type IsUnknown<T> = [unknown] extends [T]
 
 export type ValueToResponseSchema<Value> = ExtractErrorFromHandle<Value> &
 	(Extract200<Value> extends infer R200
-		? undefined extends R200
+		? Exclude<R200, undefined | void> extends never
 			? {}
 			: IsNever<R200> extends true
 				? {}


deadlinecode:
Return type `something | undefined` is now actually typed as `something | undefined | null` but i think this comes from elysia because of the json stringification and should be correct like this?

Issue #1752, opened 2026-02-23
Route specificity not working when using .use() or .group() with multiple Elysia instances


abhi-slash-git:
## Description

When composing routes from multiple Elysia instances using `.use()` or `.group()`, dynamic routes (like `/:id`) incorrectly match before static routes (like `/types`), even when the static routes are registered first.

Within a **single** Elysia instance, route specificity works correctly - `/types` is matched before `/:id` regardless of registration order. However, when routes are split across multiple instances and composed with `.use()` or `.group()`, this specificity is lost.

## Reproduction

```typescript
import { Elysia } from "elysia";

// ❌ BROKEN: Using separate instances with .use()
const typesRoutes = new Elysia({ prefix: "/types" })
  .get("/", () => "List all types")
  .get("/:id", ({ params }) => `Get type: ${params.id}`);

const itemsRoutes = new Elysia()
  .get("/", () => "List all items")
  .get("/:id", ({ params }) => `Get item: ${params.id}`);

const brokenModule = new Elysia({ prefix: "/items" })
  .use(typesRoutes)  // Static prefix - should match first
  .use(itemsRoutes); // Has /:id catch-all

// ❌ ALSO BROKEN: Using .group()
const typesRoutesNoPrefix = new Elysia()
  .get("/", () => "List all types")
  .get("/:id", ({ params }) => `Get type: ${params.id}`);

const brokenModuleWithGroup = new Elysia({ prefix: "/items" })
  .group("/types", (app) => app.use(typesRoutesNoPrefix))
  .use(itemsRoutes);

// ✅ WORKS: All routes in single instance
const workingModule = new Elysia({ prefix: "/items" })
  .get("/:id", ({ params }) => `Get item: ${params.id}`)
  .get("/", () => "List all items")
  .get("/types", () => "List all types")
  .get("/types/:id", ({ params }) => `Get type: ${params.id}`);

const app = new Elysia()
  // .use(brokenModule)        // GET /items/types returns "Get item: types" ❌
  // .use(brokenModuleWithGroup) // GET /items/types returns "Get item: types" ❌
  .use(workingModule)           // GET /items/types returns "List all types" ✅
  .listen(3000);

console.log("Test: curl http://localhost:3000/items/types");
```

## Expected Behavior

`GET /items/types` should return `"List all types"` regardless of whether routes are defined in a single instance or composed from multiple instances.

## Actual Behavior

- **Single instance**: Works correctly - `/types` matches before `/:id`
[...]

PR #1769, opened 2026-02-27
fix(#1752): specific root '/' path should not be removed

Teyik0:
fix related to -> https://github.com/elysiajs/elysia/issues/1752#issuecomment-3975075248

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed path normalization to correctly preserve root paths while maintaining expected behavior for other paths in the adapter.


SaltyAom:
Thanks for the PR

But somehow this was patched with 1ecae0a596b09093aaaedda5429dd32906794470

Issue #1741, opened 2026-02-17
Returning a `ReadableStream` results in 100% CPU and >100x latency


brettwillis:
### What version of Elysia is running?

1.4.25

### What platform is your computer?

Linux 6.12.67-linuxkit aarch64 unknown

### What environment are you using

1.3.9+cf6cdbbba

### Are you using dynamic mode?

No

### What steps can reproduce the bug?

```js
app.get('/test', async () => {

const file = Bun.file('some ~100MB file');

// Wrap the stream in another ReadableStream
// perhaps we are concatenating streams or whatever
const body = new ReadableStream({
  async start(controller) {
    const reader = file.stream().getReader();
    try {
      while (true) {
        const { done, value } = await reader.read();
        if (done)
          break;
        controller.enqueue(value);
      }
      controller.close();
    } catch (err) {
      controller.error(err);
    } finally {
      reader.releaseLock();
    }
  },
});

// Returning the stream uses 100% for several minutes
return body;

// Returning the same stream wrapped in a Response servers the stream in a fraction of a second
//return new Response(body);

});


```

### What is the expected behavior?

Returning a ReadableStream directly should have similar performance characteristics to returning a Response that wraps the stream.

### What do you see instead?

The server maxes out 1 CPU for several minutes when serving the ReadableStream directly.

PR #1772, opened 2026-03-01
fix: direct ReadableStream perf blow-up (issue #1741)

aymaneallaoui:
Fixes [1741](https://github.com/elysiajs/elysia/issues/1741)

Direct `ReadableStream` responses were super slow cuz binary chunks were getting JSON.stringify instead of streamed as bytes.

### Changes
- treat binary chunks (`Blob`/`ArrayBuffer`/typed views) as raw bytes
- keep SSE + object stream behavior the same
- add regression tests for binary stream cases

### Result
Direct `ReadableStream` perf is now in the same range as `new Response(body)` on large real-world streams, without breaking existing stream behavior.

### some perf metrics
- before: direct 100MB stream could take ~26s (100% CPU vibe mentioned on issue), while wrapped response was ~0.2s
- after: direct and wrapped are now in the same ballpark (~0.14s vs ~0.13s in local checks)

Also did randomized binary fuzz-style checks and concurrent large-stream stress runs to make sure this doesn’t regress under load.

ps: i tested all of it on same machine so yup

## Release Notes

* **Improvements**
  * Enhanced binary data streaming support with improved handling of various binary formats (Uint8Array, Blob, ArrayBuffer, and buffer views).
  * Optimized stream initialization and chunk processing for better reliability and consistency across SSE, binary, and JSON data types.

* **Tests**
  * Expanded streaming test coverage with validation of binary data handling and mixed binary source scenarios.

### DIFF
--- a/src/adapter/utils.ts
+++ b/src/adapter/utils.ts
@@ -147,6 +147,44 @@ interface CreateHandlerParameter {
 
 const allowRapidStream = env.ELYSIA_RAPID_STREAM === 'true'
 
+const isPromiseLike = (value: unknown): value is PromiseLike<unknown> =>
+	typeof (value as PromiseLike<unknown>)?.then === 'function'
+
+const isBinaryChunk = (
+	chunk: unknown
+): chunk is Blob | ArrayBuffer | ArrayBufferView =>
+	chunk instanceof Blob ||
+	chunk instanceof ArrayBuffer ||
+	ArrayBuffer.isView(chunk)
+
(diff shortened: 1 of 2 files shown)


SaltyAom:
due to a merge conflict from https://github.com/elysiajs/elysia/pull/1803, I have rewrite your code to merge with backpressure-based stream and include the test cases and mentioned in CHANGELOG.md

Thanks for your contribution

PR #1786, opened 2026-03-07
fix: default idleTimeout should be 10s to match documentation

### LINKED ISSUE #1770: Default `idleTimeout` is `30s` instead of documented `10s` inherited from Bun
### Summary

The documentation states that Elysia’s default timeout is inherited from Bun’s default of **10 seconds**:

https://elysiajs.com/patterns/configuration

However, since this commit that was released with `v1.3.9`: https://github.com/elysiajs/elysia/commit/3b3700abc1d4cc3252399ae554e6530460a049b3, the actual default timeout appears to be **30 seconds**.

This creates a discrepancy between the documented and actual behavior.

As a side-note, it might be worth mentioning that the maximum value that Bun supports today is `255s`: https://github.com/oven-sh/bun/issues/15589

### Suggested Fix

Either:

- Update the documentation to reflect the **30s** default timeout, or  
- Revert the change in the referenced commit to restore the documented **10s** behavior.

ssing2:
Fixes #1770

## Problem
The documentation states that Elysia's default `idleTimeout` is **10 seconds** (inherited from Bun), but the code was hardcoded to **30 seconds**.

Reference: https://elysiajs.com/patterns/configuration

## Solution
Changed the default `idleTimeout` from 30 to 10 seconds in two locations:
- `src/adapter/bun/index.ts` line 357
- `src/adapter/bun/index.ts` line 371

This now matches the documented behavior.

## Changes
- Modified default `idleTimeout` from 30 to 10 seconds
- Aligns code with documentation
- Consistent with Bun's default timeout behavior

## Summary by CodeRabbit

* **Chores**
  * Optimized server connection handling to improve resource efficiency.

### DIFF
--- a/src/adapter/bun/index.ts
+++ b/src/adapter/bun/index.ts
@@ -354,7 +354,7 @@ export const BunAdapter: ElysiaAdapter = {
 					? ({
 							development: !isProduction,
 							reusePort: true,
-							idleTimeout: 30,
+							idleTimeout: 10,
 							...(app.config.serve || {}),
 							...(options || {}),
 							routes,
@@ -368,7 +368,7 @@ export const BunAdapter: ElysiaAdapter = {
 					: ({
 							development: !isProduction,
 							reusePort: true,
-							idleTimeout: 30,
+							idleTimeout: 10,
 							...(app.config.serve || {}),
 							routes,
 							websocket: {


SaltyAom:
no, we will keep the 30 seconds

Will fix in the doc

PR #1798, opened 2026-03-10
fix: use resolve headers in Context.headers fallback

MegaManSec:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed type checking for request headers configuration to correctly validate header definitions.


themavik (commented):
Swapping the `{} extends` guard from `Singleton['resolve']['query']` to `headers` matches the `headers` branch below — that was a straight typing bug. clean fix.

PR #1799, opened 2026-03-10
:tada: feat: specialize response mapping via AOT for known schema types

DenisCDev:
## Summary

When a route declares a response schema (e.g. `response: t.Object({...})`), the return type is already known at compile time. Yet after validation, the response still goes through the generic `mapResponse`/`mapCompactResponse` chain which performs 10+ type checks (`constructor.name` switch, `instanceof` for Response, Blob, ReadableStream, Error, Promise, etc.).

This PR extends the existing Sucrose AOT compilation to the **response path** — generating specialized inline response code that bypasses the generic dispatch when the schema Kind is known.

### What changed

- **`getResponseSchemaKind()`** in `compose.ts` — detects TypeBox schema Kind (Object, Array, String, Number, Boolean) for single-status-code response schemas
- **`specializedResponse()`** adapter method in `adapter/types.ts` — lets each adapter generate optimized inline code per type
- **Bun adapter** — uses `Response.json()` for Object/Array (native fast path), `new Response()` for primitives
- **Web Standard adapter** — uses `JSON.stringify()` + content-type header for Object/Array
- **`mapResponse` closure** in `compose.ts` — emits a type guard with fast path + fallback to generic

### Why it helps

- Eliminates 10+ conditional branches on the hot path for typed routes
- Creates monomorphic call sites that V8/JSC can optimize with inline caches (the generic path is megamorphic)
- Completes the partial evaluation pattern Elysia already applies to the request path (Sucrose) — now the response path is specialized too

### Safety

Every specialized path has a runtime type guard that falls back to the generic `mapResponse` if the value doesn't match:

```js
r !== null && r !== undefined && r.constructor === Object
  ? Response.json(r)        // fast path
  : mapResponse(r, c.set)   // fallback
```

Specialization is **disabled** when:
- No response schema declared
- Multiple status codes in schema
- Union/Intersect types (Kind not whitelisted)
- `mapResponse` hooks are present (could change type)
- `maybeStream` is true (generators/streams)
- Standard Schema provider (no Kind property)

### Test plan

- [x] All 1508 existing tests pass (`bun test`) — zero regressions
- [ ] Benchmark typed vs untyped routes with `autocannon` / `bun bench`
- [ ] Inspect generated `fnLiteral` to verify inline `Response.json()` appears

## Summary by CodeRabbit

* **Refactor**
  * Introduced optimized fast-path response handling across adapters and core composition to return JSON, text, number, or boolean responses more directly, improving runtime routing and performance while preserving existing behavior and fallbacks.
  * Adapter interfaces extended to optionally support these specialized response paths.

### DIFF
--- a/src/adapter/bun/index.ts
+++ b/src/adapter/bun/index.ts
@@ -220,7 +220,28 @@ export const BunAdapter: ElysiaAdapter = {
 			? 'c.headers=c.request.headers.toJSON()\n'
 			: 'c.headers={}\n' +
 				'for(const [k,v] of c.request.headers.entries())' +
-				'c.headers[k]=v\n'
+				'c.headers[k]=v\n',
+		specializedResponse(kind, r, hasSet, saveResponse) {
+			// Bun's Response.json() is faster than new Response(JSON.stringify())
+			switch (kind) {
+				case 'Object':
+				case 'Array':
+					return hasSet
+						? `Response.json(${saveResponse}${r},c.set)`
+						: `Response.json(${saveResponse}${r})`
+				case 'String':
+					return hasSet
+						? `new Response(${saveResponse}${r},c.set)`
+						: `new Response(${saveResponse}${r})`
+				case 'Number':
+				case 'Boolean':
+					return hasSet
+						? `new Response(${saveResponse}''+${r},c.set)`
+						: `new Response(${saveResponse}''+${r})`
+				default:
+					return undefined
+			}
+		}
 	},
 	listen(app) {
 		return (options, callback) => {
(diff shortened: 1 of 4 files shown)


DenisCDev:
The generic `mapResponse` handler for Number/Boolean in web-standard also does `new Response(response.toString(), set)` without setting content-type ([handler.ts L110-115](https://github.com/elysiajs/elysia/blob/main/src/adapter/web-standard/handler.ts#L110-L115)). The specialized path matches the existing behavior exactly, so no change needed here.

DenisCDev:
Closing this — five months without maintainer interest is a clear enough answer, and leaving it open only adds to the queue.

The change is still valid if it is ever wanted: it extends the existing Sucrose AOT compilation to the response path, so routes that declare a `response` schema skip the generic `mapResponse` dispatch and its ten-plus type checks. Happy to rebase and reopen if that direction is interesting.

PR #1802, opened 2026-03-12
Handle Range header for File/Blob responses

### LINKED ISSUE #1790: `Range` header is ignored

elysia@1.4.26

N/A

Bun 1.3.9

No

```
echo "12345" > foobar.txt
```

```ts
import { Elysia } from "elysia"
new Elysia().get("/file", () => Bun.file("foobar.txt")).listen(3000)
```

```
curl -H "Range: bytes=3-" http://localhost:3000/file
```

Should slice text file and return data from byte 3 and onward (`345`)

### What do you see instead?

Returns the whole file starting with byte 0 — content-range: bytes 0-5/6 (`12345`)

### Additional information

_No response_

N/A

MarcelOlsen:
this closes #1790

When returning a Blob or File response, the Range request header was ignored, resulting in clients always receiving the full file.

The bug had two layers. First, `handleFile` never received the `request` object, preventing it from reading the header. Additionally, `compose.ts` only forwarded `c.request` to the response mappers when `maybeStream` was `true`. This meant that for any plain file route, the request was dropped before reaching `handleFile`.

Changes:

* Removed the `maybeStream` guard in `compose.ts`, ensuring `c.request` is always forwarded to response mappers.
* `handleFile` now accepts an optional request parameter and handles range requests by slicing the blob and returning a `206` status with the appropriate `Content-Range` and `Content-Length` headers.
* Supports all three RFC 7233 range forms (`bytes=start-end`, `bytes=start-`, `bytes=-suffix`), clamps oversized end positions per specification, and returns a `416` status when the start is out of bounds.
* Threaded the request through all `mapResponse` and `mapCompactResponse` call sites in both the bun and web-standard adapters.

A regression test was added to cover all the above cases.

## Summary by CodeRabbit

* **New Features**
  * File responses now honor HTTP Range requests: returns 206 with accurate Content-Range/Content-Length for valid ranges, and 416 with Content-Range */size for invalid or unsatisfiable ranges; supports open-ended, suffix, and clamped ranges while preserving other headers.

* **Bug Fixes**
  * File/stream handling now receives request context consistently, improving correctness for request-dependent file responses.

* **Tests**
  * Added regression tests covering full, open-ended, bounded, suffix, out-of-range, invalid, and multi-range scenarios.

### DIFF
--- a/src/adapter/utils.ts
+++ b/src/adapter/utils.ts
@@ -16,13 +16,16 @@ export const handleFile = (
 	const size = response.size
 
 	const rangeHeader = request?.headers.get('range')
-	if (rangeHeader && size) {
+	if (rangeHeader) {
 		const match = /bytes=(\d*)-(\d*)/.exec(rangeHeader)
 		if (match) {
 			if (!match[1] && !match[2])
 				return new Response(null, {
 					status: 416,
-					headers: { 'content-range': `bytes */${size}` }
+					headers: mergeHeaders(
+						new Headers({ 'content-range': `bytes */${size}` }),
+						set?.headers ?? {}
+					)
 				})
 
 			let start: number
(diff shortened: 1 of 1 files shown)


MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

PR #1803, opened 2026-03-12
Replace eager start() loop with pull()-based backpressure

### LINKED ISSUE #1801: SSE async generator has zero backpressure OOM on large streams

`elysia@1.4.27`

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 x86_64

Bun 1.3.10

No

`createStreamHandler` in `adapter/utils.mjs` wraps async generators in `new ReadableStream({ start() })`, which eagerly drains the entire generator without waiting for the consumer to read. This causes OOM when streaming large datasets via SSE.

In `adapter/utils.mjs` (~line 91):

```js
new ReadableStream({
  async start(controller) {
    for await (const chunk of generator) {
      controller.enqueue(format(...));
    }
    controller.close();
  }
})
```

With `start()`, the entire generator is consumed as fast as the source can produce, buffering everything into the ReadableStream's internal queue. The TCP socket already provides backpressure (when the client is slow, `write()` returns false), but Elysia never sees that signal because it buffers everything in JS memory before the data reaches the socket.

For SSE streams specifically, even the `setTimeout(0)` yield that exists for non-SSE streams is skipped (~line 134: `!allowRapidStream && isBrowser && !isSSE`).

A backpressure-aware implementation would use `pull()`:

```js
new ReadableStream({
  async pull(controller) {
    const { value, done } = await generator.next();
    if (done) { controller.close(); return; }
    controller.enqueue(format(value));
  }
})
```

**data-source.ts** — run with `bun run data-source.ts`:
```ts
Bun.serve({
  port: 9999,
  fetch() {
    let i = 0;
    return new Response(
      new ReadableStream({
        pull(controller) {
          if (i >= 40_000_000) { controller.close(); return; }
          let chunk = "";
          for (const end = Math.min(i + 100, 40_000_000); i < end; i++)
            chunk += JSON.stringify({ i, pad: "x".repeat(200) }) + "\n";
          controller.enqueue(new TextEncoder().encode(chunk));
        },
      }),
    );
  },
});
```

**repro-elysia.ts** — run with `bun run repro-elysia.ts`:
```ts
import { Elysia, sse } from "elysia";

new Elysia()
  .get("/stream", async function* () {
    const res = await fetch("http://localhost:9999/");
    const reader = res.body!.getReader();
    const decoder = new TextDecoder();
    let buf = "";
    while (true) {
      const { value, done } = await reader.read();
      if (done) break;
      buf += decoder.decode(value, { stream: true });
      const lines = buf.split("\n");
      buf = lines.pop()!;
      for (const line of lines) if (line) yield sse({ data: line });
    }
  })
  .listen(3002);

setInterval(() => console.log(`RSS: ${(process.memoryUsage.rss() / 1024 / 1024).toFixed(0)} MB`), 2000);
```

```bash
curl -sN --limit-rate 1M http://localhost:3002/stream > /dev/null
# Server gets OOM-killed within seconds
```

Memory should stay bounded regardless of result set size. The generator should only advance when the downstream TCP socket is ready for more data.

| Server | Result |
|--------|--------|
| **Elysia SSE generator** | **OOM-killed at 14.4 GB RSS** |
| Node.js `http` + `stream.pipeline()` | 87 MB RSS (stable) |

### What do you see instead?

Memory grows to match the full upstream response size. The process gets OOM-killed within seconds, it goes from 78 MB to 14.4 GB so fast it doesn't even print a single memory log line (2s interval).

### Additional information

- #1768
- #1741

Yes

MarcelOlsen:
this resolves #1801

## Problem

`createStreamHandler` uses `new ReadableStream({ async start() })`, which eagerly drains the entire generator into the stream's internal queue without waiting for the consumer. When a generator produces data faster than the consumer can read, such as when SSE proxies a large upstream, memory usage can grow unbounded. This issue was reported as an out-of-memory (OOM) error, escalating from 78 MB to 14.4 GB in seconds.

## Root Cause

The Streams specification calls `start()` only once, immediately, before any reads occur. Placing a `for await...of` loop there does not create backpressure; instead, the loop races ahead of the consumer and fills the internal queue without limit.

The solution is to use `pull()`, which the runtime calls **only when the consumer is ready for the next chunk**, resulting in one invocation per read.

## Solution

Extract an explicit `AsyncIterator` before constructing the `ReadableStream`, then divide the single `async start()` into three callbacks:

-  `start()` (sync) — registers the abort handler and enqueues the already-extracted initial value
-  `async pull()` — advances the iterator by one step on demand
-  `cancel()` — calls `iterator.return()` to clean up the generator

This approach also removes the `isBrowser` + `setTimeout(0)` workaround introduced in #741. That hack was necessary because back-to-back `enqueue()` calls in the same microtask caused chunks to merge in the consumer's buffer. With `pull()`, each enqueue occurs in a separate microtask driven by a consumer read, preventing the original issue from recurring.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming/backpressure behavior to avoid eager consumption of producers, with more reliable abort/cancel handling and consistent SSE/JSON chunk formatting.

* **Tests**
  * Added regression tests validating backpressure for proxied SSE streams and lazy generators to ensure producers are not drained faster than consumers.

### DIFF
--- a/test/response/stream.test.ts
+++ b/test/response/stream.test.ts
@@ -641,6 +641,42 @@ describe('Stream', () => {
 		])
 	})
 
+	// Regression: proxying a large upstream SSE stream caused OOM (#1801).
+	// The upstream generator must not be drained ahead of the slow consumer.
+	it('does not buffer unboundedly when proxying a slow-consuming SSE stream', async () => {
+		const TOTAL = 50
+		let produced = 0
+
+		const upstream = new Elysia().get('/', async function* () {
+			for (let i = 0; i < TOTAL; i++) {
+				produced++
+				yield sse(`message ${i}`)
+			}
+		})
+
+		// Simulate the OOM scenario: one Elysia instance re-streams another's
+		// SSE response while the consumer reads slowly.
+		const proxy = new Elysia().get('/', () =>
+			upstream.handle(new Request('http://e.ly'))
+		)
+
+		const response = await proxy.handle(req('/'))
+		const reader = response.body!.getReader()
+
+		// Slow consumer: read 3 chunks with pauses between each
+		await reader.read()
+		await Bun.sleep(20)
+		await reader.read()
+		await Bun.sleep(20)
+		await reader.read()
+
+		// The upstream generator must not have raced ahead and produced all
+		// TOTAL chunks. A small prefetch buffer is fine, but nowhere near 50.
+		expect(produced).toBeLessThan(TOTAL)
+
+		reader.cancel()
+	})
+
 	// Regression test for https://github.com/elysiajs/elysia/issues/1801
 	// The generator must not be drained ahead of the consumer (backpressure).
 	it('does not eagerly drain generator ahead of consumer', async () => {


MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

SaltyAom:
this is fire

PR #1805, opened 2026-03-13
Fix dynamic imports inside `.guard()` not registering routes

### LINKED ISSUE #1774: Bug: Routes from dynamic import are not registered when used inside .guard()

1.4.27

Linux 6.19.5-3-cachyos x86_64 unknown

1.3.10

no

In this example, when using dynamic imports inside a .guard(), the routes are not registered

plugin-a.ts :
```ts

const pluginA = new Elysia()
	.get('/a', () => 'Hello from Plugin A!');

export default pluginA;
```

main.ts : 
```ts
import { openapi } from '@elysiajs/openapi';

new Elysia()
	.use(openapi())
	.guard({}, (app) => app.use(import('./plugin-a')))
	.listen(3000, () => {
		console.log('Server is running on http://localhost:3000');
	});
```

I expect the routes to be registered even behind a guard

### What do you see instead?

no routes 😢 

### Additional information

_No response_

_No response_

MarcelOlsen:
Fixes #1774

## Root cause

`guard()` creates a sandbox `Elysia` instance, passes it to the user's callback, then synchronously iterates `instance.router.history` to re-register each route on the parent with the guard hooks applied.

When the callback uses a dynamic import — `app.use(import('./plugin'))` — `use()` recognises the `Promise` and defers resolution into `instance.promisedModules`, returning synchronously. By the time the `forEach` runs, `router.history` is empty and the routes are never registered. The sandbox's `promisedModules` are also never propagated to the parent, so the promises are abandoned.

## Fix

After the synchronous route iteration, check whether the sandbox has pending promised modules. For each one, add a corresponding promise to `this.promisedModules` that — once the dynamic import settles and its routes have been pushed into `instance.router.history` — registers the newly added routes on the parent with the guard hooks applied.

A shared `processedUntil` marker (rather than a fixed `syncCount`) ensures each callback only processes its own slice of `router.history`. Because JS is single-threaded, each `.then()` callback atomically snapshots `startIndex = processedUntil` and advances the marker before iterating — so even if multiple dynamic imports resolve in the same microtask drain, every route is registered exactly once.

## Tests

- Routes from a dynamic import inside `guard` are reachable
- Guard lifecycle hooks (`beforeHandle`) are applied to dynamically imported routes
- Multiple dynamic imports inside a single `guard` are each reachable and their shared guard hook fires exactly once per request
    

## Summary by CodeRabbit

* **New Features**
  * Support for dynamic imports inside route guards, allowing asynchronous registration of routes while preserving guard hooks and validators

* **Bug Fixes / Improvements**
  * Smoother integration between synchronous and asynchronous guard-based route registration and error/validator handling

* **Tests**
  * Added tests validating dynamic import route registration and guard hook invocation

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4685,6 +4685,77 @@ export default class Elysia<
 			}
 		)
 
+		// Handle dynamic imports (Promises) used inside guard callback
+		if (instance.promisedModules.size > 0) {
+			const syncCount = instance.router.history.length
+
+			for (const promise of instance.promisedModules.promises) {
+				this.promisedModules.add(
+					promise.then(() => {
+						const {
+							body,
+							headers,
+							query,
+							params,
+							cookie,
+							response,
+							...guardHook
+						} = hook
+
+						const hasStandaloneSchema =
+							body || headers || query || params || cookie || response
+
+						for (
+							let i = syncCount;
+							i < instance.router.history.length;
+							i++
+						) {
+							const {
+								method,
+								path,
+								handler,
+								hooks: localHook
+							} = instance.router.history[i]
+
+							this.add(
+								method,
+								path,
+								handler,
+								mergeHook(guardHook as AnyLocalHook, {
+									...((localHook || {}) as AnyLocalHook),
+									error: !localHook.error
+										? sandbox.event.error
+										: Array.isArray(localHook.error)
+											? [
+													...(localHook.error ?? []),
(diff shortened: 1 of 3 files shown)


MarcelOlsen:
@coderabbitai review

MarcelOlsen:
@coderabbitai review

Necrelox:
thank you ❤️

PR #1810, opened 2026-03-17
fix(error): use explicit undefined check in status() fallback to support null body

### LINKED ISSUE #1738: Returning explicit status codes interferes with response validation

1.4.1

1.3.7

No

Create a new app that defines some response types, and then return one with an explicit status code. Like in this snippet:

```ts
import { Elysia, StatusMap, t } from "elysia";
import { register } from "../modules/auth";
import { errorResponseSchema, internalErrorResponse } from "@lib/http";

new Elysia().post(
    "/register",
    async ({ body, status }) => {
      const result = await register({
        email: body.email,
        password: body.password,
        name: body.name,
      })

      if (result.isSuccess) {
        // ...
        return status(StatusMap.Created, null);  // <--- Causes response validation to fail
      }

      switch (result.error) {
        case "DUPLICATE_EMAIL":
          return status(StatusMap["Bad Request"], {
            errorMessage:
              "This email is already registered, please login or try another email",
            errorCode: "DUPLICATE_EMAIL",
            statusCode: StatusMap["Bad Request"],
          });
        default:
          return status(
            StatusMap["Internal Server Error"],
            internalErrorResponse,
          );
      }
    },
    {
      body: t.Object({
        name: t.String(),
        email: t.String({ format: "email" }),
        password: t.String({ minLength: 8, maxLength: 255 }),
      }),
      response: {
        [StatusMap["Bad Request"]]: errorResponseSchema,
        [StatusMap["Internal Server Error"]]: errorResponseSchema,
        [StatusMap.Created]: t.Null(),
      },
    },
  )
```

When I execute the request, and the logic is correctly executed, the response validation returns a 422 for this return statement: `return status(StatusMap["Created"], null);`

The resulting validation error: 
```json
{
    "type": "response",
    "validator": {
      "schema": {
        "type": "null"
      },
      "references": [],
      "code": "return function check(value) {\n  return (\n    (value === null)\n  )\n}",
      "hasTransform": false,
      "provider": "typebox",
      "hasAdditionalProperties": false,
      "hasDefault": false,
      "isOptional": false
    },
    "value": "Created",
    "allowUnsafeValidationDetails": false,
    "code": "VALIDATION",
    "status": 422,
    "valueError": {
      "type": 35,
      "schema": {
        "type": "null"
      },
      "path": "",
      "value": "Created",
      "message": "Expected null",
      "errors": []
    },
    "expected": null
}
```

As we can see it says it receives "Created" which is the string value of the status code we return, where it expected "null". It seems the status code interferes with the response validation?

It should not try to validate the status code, numerical or not, as part of the response body.

### What do you see instead?

It interprets the status code, or the string version "Created" as being part of the response body.

### Additional information

I believe [this issue](https://github.com/elysiajs/elysia/issues/1480) might be related , however I've provided a snippet and more information.

_No response_

alikia2x:
## Summary

- Fixes #1738 

## Problem

Currently, the `status(code, response)` helper uses the nullish coalescing operator (`??`) to provide a default response body (the status text, e.g., "Created") when the second argument is missing.

https://github.com/elysiajs/elysia/blob/56310be9617b826f862c985eae95ae823d95f097/src/error.ts#L79-L106  

However, this causes an unintended side effect: explicitly passing `null` as the response body is treated the same as passing `undefined`.

This behavior interferes with response validation schemas where `null` is expected body type.

## Solution

This PR changes the fallback logic to strictly check for `undefined` instead of using `??`.

**Changes in `src/error.ts`:**
```typescript
// Before
const res = response ?? (code in InvertedStatusMap ? InvertedStatusMap[code] : code);

// After
const res = response === undefined 
  ? (code in InvertedStatusMap ? InvertedStatusMap[code] : code) 
  : response;
```

## Result

`status(201, null)` now correctly returns `null` (empty body), allowing response validators to pass when the schema expects `null`.

## Test Coverage
- Verified that existing test cases still pass.
- Added two new test cases to verify that `status(201, null)` correctly returns an empty body and passes response validation against a `t.Null()` schema.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of null response bodies in status mapping to ensure consistent behavior across response types.

* **Tests**
  * Added test coverage for null and empty body response scenarios.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -89,11 +89,12 @@ export class ElysiaCustomStatusResponse<
 
 	constructor(code: Code, response: T) {
 		const res =
-			response ??
-			(code in InvertedStatusMap
-				? // @ts-expect-error Always correct
-					InvertedStatusMap[code]
-				: code)
+			response === undefined
+				? code in InvertedStatusMap
+					? // @ts-expect-error Always correct
+						InvertedStatusMap[code]
+					: code
+				: response
 
 		// @ts-ignore Trust me bro
 		this.code = StatusMap[code] ?? code
--- a/test/core/status.test.ts
+++ b/test/core/status.test.ts
@@ -121,4 +121,13 @@ describe('Status', () => {
 		expect(response.status).toBe(308)
 		expect(await response.text()).toBe('')
 	})
+
+	it('return empty body when body is explicitly set to null', async () => {
+		const app = new Elysia().get('/', ({ status }) => status(201, null))
+
+		const response = await app.handle(req('/'))
+
+		expect(response.status).toBe(201)
+		expect(await response.arrayBuffer()).toHaveLength(0);
+	})
 })
--- a/test/validator/response.test.ts
+++ b/test/validator/response.test.ts
@@ -582,4 +582,16 @@ describe('Response Validator', () => {
 
 		expect(status).resolves.toBe(200)
 	})
+
+	it('handle empty body', () => {
+		const app = new Elysia().get('/', ({ status }) => status(201, null), {
+			response: {
+				201: t.Null()
+			}
+		})
+
+		const status = app.handle(req('/')).then((x) => x.status)
+
+		expect(status).resolves.toBe(201)
+	})
 })


themavik (commented):
Using `response === undefined` stops `status(201, null)` from falling back to `InvertedStatusMap` / numeric `code`. clean fix; the new `status.test` covers the empty body case well.

PR #1815, opened 2026-03-24
fix: preserve null/falsy body in ElysiaCustomStatusResponse

### LINKED ISSUE #1738: Returning explicit status codes interferes with response validation

eL1fe:
## Summary
- `status(201, null)` was replacing `null` with the string `"Created"`, causing response validation to fail with 422
- Root cause: nullish coalescing (`??`) treats `null` as nullish and substitutes the status name
- Changed to explicit `undefined` check so only truly omitted bodies get the fallback

## Problem
```typescript
new Elysia()
  .post('/test', ({ status }) => {
    return status(201, null) // 422 error: validates "Created" string instead of null
  }, {
    response: { 201: t.Null() }
  })
```

The `??` operator in `ElysiaCustomStatusResponse` constructor replaced `null`, `0`, `false`, and `""` with the status name string (e.g. `"Created"`).

## Changes
- `src/error.ts`: replace `response ?? fallback` with `response !== undefined ? response : fallback`

## Test plan
- [x] `null` body preserved correctly
- [x] `undefined` body falls back to status name (existing behavior)
- [x] `0`, `false`, `""` bodies preserved correctly
- [x] All existing tests pass (461 pass, 0 fail across core + validator)

Fixes #1738

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed custom status response handling to properly treat explicitly passed null values instead of using default fallbacks.

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -89,11 +89,12 @@ export class ElysiaCustomStatusResponse<
 
 	constructor(code: Code, response: T) {
 		const res =
-			response ??
-			(code in InvertedStatusMap
-				? // @ts-expect-error Always correct
-					InvertedStatusMap[code]
-				: code)
+			response !== undefined
+				? response
+				: code in InvertedStatusMap
+					? // @ts-expect-error Always correct
+						InvertedStatusMap[code]
+					: code
 
 		// @ts-ignore Trust me bro
 		this.code = StatusMap[code] ?? code


mikededo:
Duplicate of #1810, right?

eL1fe:
Yeah, looks like #1810 addresses the same issue. Happy to close this one in favor of it.

Issue #1277, opened 2025-07-05
Using Cloudflare workers and returning a 204, returns a body which is unacceptable by Cloudflare


jasperdunn:
### What version of Elysia is running?

elysia@1.3.5

### What platform is your computer?

Darwin 24.5.0 arm64 arm

### What steps can reproduce the bug?

```ts
return status(204, null);
```

```ts
return new Response(null, { status: 204 });
```

Returning a 204 in either way causes this Cloudflare error:
> Constructing a Response with a null body status (204) and a non-null, zero-length body. This is technically incorrect, and we recommend you update your code to explicitly pass in a `null` body, e.g. `new Response(null, { status: 204, ... })`. (We continue to allow the zero-length body behavior because it was previously the only way to construct a Response with a null body status. This behavior may change in the future.)

A similar issue was raised for Remix here: https://github.com/remix-run/remix/issues/553

### What is the expected behavior?

Pass `null` to Cloudflare. I think Elysia is passing along a string at the moment, not sure how to replicate this.

### Additional information

Can we also change the types so that `status(204)` can be written without requiring the second parameter?

### Have you try removing the `node_modules` and `bun.lockb` and try again yet?

Yes

gewoonwoutje:
When trying this in Deno using Elysia v1.3.13, I get a similar 500 error response when using `status(204)`, but not when creating the `new Response(null, { status: 204})` directly.

<img width="804" height="673" alt="Image" src="https://github.com/user-attachments/assets/af60b98b-679a-40a5-aa4a-e8f5540df625" />

Code snippet used:

```ts
import { assertEquals } from '@std/assert';
import { Elysia } from 'elysia';

Deno.test('status(204) has empty body', async () => {
	const app = new Elysia().get('/no-content', ({ status }) => status(204));

	const response = await app.handle(new Request('http://localhost/no-content'));
	const text = await response.text();

	console.error('Text: ', text);
	console.error('Status: ', response.status);

	assertEquals(text, '');
	assertEquals(response.status, 204);
});

```

I believe the Response is constructed using an empty string, which is not allowed in web standard:

<img width="768" height="134" alt="Image" src="https://github.com/user-attachments/assets/af2f1667-73d5-4a7e-a1b5-af22cbbc753a" />

sparten11740:
we're facing the same issue on Vercel with `elysia@1.4.13` and `@elysiajs/node@1.4.1` when using `status(204)`

[...]

PR #1833, opened 2026-03-31
fix: empty-body status codes send body when using string status names

alexkahndev:
## Summary

- `ElysiaCustomStatusResponse` constructor checks `code in emptyHttpStatus`, but when using string status names like `status('No Content')`, `code` is `'No Content'` while `emptyHttpStatus` keys are numeric (`204`). The lookup always fails, so `this.response` is never cleared — resulting in a response body being sent with 204/205/304/etc.
- Fixed by checking `this.code in emptyHttpStatus` instead, since `this.code` is already resolved to the numeric status via `StatusMap`.
- Also changed `mapResponse`'s `case undefined` to use `new Response(null, set)` instead of `new Response('', set)` in both the bun and web-standard adapters, since HTTP spec requires no body for these status codes and an empty string still produces a `ReadableStream` body.

## Test plan

- [x] Added `test/response/no-content-status.test.ts` with 3 tests covering 204 (no args), 204 (with message arg), and 205
- [x] Verified all 3 tests fail before the fix and pass after
- [x] Full test suite passes (1528/1528)

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized empty/falsy response handling so no-content statuses (204, 205) return null bodies across environments.
  * Use the resolved HTTP status when deciding to omit a response body for status-based responses.

* **Tests**
  * Added tests verifying 204/205 responses produce null bodies and correct HTTP statuses.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elysiajs/elysia/pull/1833?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/adapter/bun/handler.ts
+++ b/src/adapter/bun/handler.ts
@@ -56,7 +56,7 @@ export const mapResponse = (
 				)
 
 			case undefined:
-				if (!response) return new Response('', set as any)
+				if (!response) return new Response(null, set as any)
 
 				return Response.json(response, set as any)
 
--- a/src/adapter/bun/handler.ts
+++ b/src/adapter/bun/handler.ts
@@ -307,7 +307,7 @@ export const mapEarlyResponse = (
 				)
 
 			case undefined:
-				if (!response) return new Response('')
+				if (!response) return new Response(null)
 
 				return Response.json(response)
 
@@ -423,7 +423,7 @@ export const mapCompactResponse = (
 			)
 
 		case undefined:
-			if (!response) return new Response('')
+			if (!response) return new Response(null)
 
 			return Response.json(response)
 
--- a/src/adapter/bun/handler.ts
+++ b/src/adapter/bun/handler.ts
@@ -307,7 +307,7 @@ export const mapEarlyResponse = (
 				)
 
 			case undefined:
-				if (!response) return new Response(null)
+				if (!response) return new Response()
 
 				return Response.json(response)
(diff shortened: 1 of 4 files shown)


TheNumberOne:
I've been running into this same issue as well in a NodeJS environment + web standard adapter.

TheNumberOne (approved):
My only feedback is that the default value for the first argument is `null`. It might be clearer to omit the null in calls that have only one argument. It might also be clearer to use `undefined` instead

alexkahndev:
@TheNumberOne good call — done in 26de6b89. I omitted `null` in the single-arg cases (`mapEarlyResponse`/`mapCompactResponse` in both the bun and web-standard handlers), so they're now `new Response()`.

I kept the two-arg `mapResponse` calls as `new Response(null, set as any)` since the body arg can't be omitted when an init object is passed — minor asymmetry, but deliberate.

On `undefined`: I left it as `null` for those two-arg calls rather than switching, since `null` is the fetch spec's explicit "no body" sentinel, whereas `undefined` just relies on default-param fall-through. Happy to revisit if you feel strongly though.

TheNumberOne:
@alexkahndev That is convincing to me!

I'm not a maintainer on this project fyi. Just also interested in this PR getting merged.

TheNumberOne:
Linky to relevant issue: https://github.com/elysiajs/elysia/issues/1277

PR #1850, opened 2026-04-12
fix: strip prototype pollution keys from FormData JSON parsing

### LINKED ISSUE #1848: [Security] Query type confusion and FormData auto-parse prototype pollution
## Summary
Two security issues found in Elysia's request parsing:

### 1. Query Type Confusion (Medium)
**File:** `src/parse-query.ts:121-239`

`parseQueryStandardSchema` (used with Zod/Valibot) auto-parses query values that look like JSON objects/arrays and splits commas into arrays. This happens **before** schema validation, so transform/derive hooks receive objects where strings were expected. Applications using `noValidate` are fully exposed.

**Example:** `?filter={"$gt":0}` is parsed as an object instead of a string.

**Fix:** Don't auto-parse query values before validation, or at minimum only do so when the schema explicitly expects non-string types.

### 2. FormData Auto-Parse Prototype Pollution (Medium)
**File:** `src/dynamic-handle.ts:126-169`

`normalizeFormValue` auto-parses any form field value starting with `{` or `[` as JSON. Parsed objects can carry `__proto__` as an enumerable property, creating prototype pollution risk if downstream code merges/spreads these objects.

**Fix:** Strip `__proto__`, `constructor`, and `prototype` keys from parsed FormData objects (similar to how cookie parsing already does this).

## Impact
- Type confusion can bypass application security logic that expects string query params
- Prototype pollution via FormData can affect application behavior

## Affected Version
Current main branch

eddieran:
## Summary

Fixes #1848 — FormData auto-parse prototype pollution.

This replaces #1849 (which was auto-closed when the fork was deleted) and addresses the gap identified in the CodeRabbit review: the nested key path parsing code was calling `JSON.parse` on intermediate values without stripping dangerous keys.

### Changes

- Added a `stripDangerousKeys` function that recursively removes `__proto__`, `constructor`, and `prototype` keys from parsed objects
- Applied `stripDangerousKeys` to **all** `JSON.parse` results in FormData processing:
  - `normalizeFormValue` in `src/dynamic-handle.ts` (single-value JSON parse and file+JSON mixed parse)
  - `parseObjectString` in `src/dynamic-handle.ts` (nested key path intermediate object parsing)
  - All three `JSON.parse` call sites in the inlined `formData` parser in `src/adapter/web-standard/index.ts` — including the nested key path `JSON.parse(existing)` that the review flagged
- Added comprehensive tests covering `__proto__`, `constructor`, `prototype` stripping at top-level, nested, and deeply nested JSON, as well as nested key path (`user.profile`) and dangerous key path (`constructor.prototype`) scenarios

### Test plan

- All 8 new tests in `test/core/formdata-prototype-pollution.test.ts` pass
- Existing `test/core/formdata.test.ts` tests continue to pass (no regressions)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved form-data and JSON parsing to strip unsafe prototype-related keys from incoming values, preventing prototype-pollution from crafted inputs.

* **Tests**
  * Added end-to-end tests validating protection against prototype pollution across single fields, nested structures, arrays, dotted/bracket paths, and deep JSON payloads.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -34,7 +34,7 @@
       "peerDependencies": {
         "@sinclair/typebox": ">= 0.34.0 < 1",
         "@types/bun": ">= 1.2.0",
-        "exact-mirror": "^0.2.6",
+        "exact-mirror": ">= 0.0.9",
         "file-type": ">= 20.0.0",
         "openapi-types": ">= 12.0.0",
         "typescript": ">= 5.0.0",
--- a/src/adapter/web-standard/index.ts
+++ b/src/adapter/web-standard/index.ts
@@ -58,6 +58,15 @@ export const WebStandardAdapter: ElysiaAdapter = {
 					`const m=k.match(/^(.+)\\[(\\d+)\\]$/);` +
 					`return m?{name:m[1],index:parseInt(m[2],10)}:null` +
 					`}\n` +
+					`const stripDangerousKeys=(o)=>{` +
+					`if(o===null||typeof o!=='object')return o;` +
+					`if(Array.isArray(o))return o.map(stripDangerousKeys);` +
+					`const r={};` +
+					`for(const k of Object.keys(o)){` +
+					`if(dangerousKeys.has(k))continue;` +
+					`r[k]=stripDangerousKeys(o[k])` +
+					`}return r` +
+					`}\n` +
 					`for(const key of form.keys()){` +
 					`if(c.body[key])continue\n` +
 					`const value=form.getAll(key)\n` +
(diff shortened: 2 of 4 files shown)


eddieran:
Thanks @coderabbitai — valid catch. In an object literal, `__proto__` is treated as a prototype setter (not an own enumerable property), so `JSON.stringify({ __proto__: {...} })` drops it entirely and the sanitizer was never actually seeing the dangerous key on the wire. I rewrote the payloads to raw JSON strings (and built the top-level `FormData` by hand for the same reason). All 8 tests still pass.

PR #1862, opened 2026-04-27
fix: encode stream chunks as Uint8Array for runtime compatibility

### LINKED ISSUE #1845: sse() helper does not work on Deno — ReadableStream enqueues raw strings instead of Uint8Array
## Description

The `sse()` helper combined with the generator pattern does not work on Deno. The SSE response returns **200 OK** with `Content-Type: text/event-stream`, but **no events are ever transmitted** to the client. The EventStream tab in browser DevTools is completely empty.

The same code works correctly on Bun.

## Root Cause

In `adapter/utils.js`, the `createStreamHandler` enqueues SSE data as **raw strings** into the `ReadableStream`:

```javascript
// adapter/utils.js:168-169 (start callback)
if (init.value.toSSE)
    controller.enqueue(init.value.toSSE());  // toSSE() returns a string

// adapter/utils.js:201-202 (pull callback)
if (chunk.toSSE)
    controller.enqueue(chunk.toSSE());  // same issue
```

The `toSSE()` method (in `utils.js`) returns a plain string like `"event: progress\ndata: {...}\n\n"`. This string is passed directly to `controller.enqueue()` without encoding.

A `ReadableStream` used as a `Response` body should contain `Uint8Array` chunks per the web standard. **Bun** is lenient and auto-encodes string chunks to bytes. **Deno** follows the spec strictly — the raw strings are silently not transmitted.

## Reproduction

```typescript
import { Elysia, sse } from 'elysia';

new Elysia()
    .get('/sse', async function* () {
        yield sse({ event: 'ping', data: { message: 'hello' } });
        yield sse({ event: 'ping', data: { message: 'world' } });
    })
    .listen(3000);
```

1. Run with `deno run --allow-net server.ts`
2. Open browser to `http://localhost:3000/sse` or use `new EventSource('/sse')`
3. **Expected:** Two `ping` events received
4. **Actual:** Response is 200 with `text/event-stream` content-type, but EventStream tab is empty. No events arrive. The `EventSource` `error` event fires due to connection close.

Works correctly when running with Bun.

## Suggested Fix

Encode the string to `Uint8Array` before enqueuing in `adapter/utils.js`:

```javascript
const encoder = new TextEncoder();

// In start callback:
if (init.value.toSSE)
    controller.enqueue(encoder.encode(init.value.toSSE()));

// In pull callback:
if (chunk.toSSE)
    controller.enqueue(encoder.encode(chunk.toSSE()));
```

## Workaround

Return a manual `Response` with a self-built `ReadableStream` instead of using the `sse()` + generator pattern:

```typescript
.get('/sse', () => {
    const encoder = new TextEncoder();
    const stream = new ReadableStream({
        async start(controller) {
            // ... produce SSE data ...
            const sseString = `event: ping\ndata: ${JSON.stringify({ message: 'hello' })}\n\n`;
            controller.enqueue(encoder.encode(sseString));
            controller.close();
        },
    });
    return new Response(stream, {
        headers: {
            'Content-Type': 'text/event-stream',
            'Cache-Control': 'no-cache',
            'Connection': 'keep-alive',
        },
    });
})
```

## Environment

- **Elysia:** 1.4.28
- **Deno:** 2.x
- **OS:** macOS (Darwin arm64)

Zelys-DFKH:
Fixes #1845.

\`ReadableStream\` controllers are required by the spec to receive \`BufferSource\` chunks (\`Uint8Array\`, \`ArrayBuffer\`, etc.). Bun accepts raw strings as a convenience extension, but spec-strict runtimes like Deno silently drop string chunks, causing generators and SSE to produce empty responses.

## Root cause

\`createStreamHandler\` in \`src/adapter/utils.ts\` enqueued strings directly in both the \`start()\` and \`pull()\` callbacks: 8 call sites in total.

## Fix

Added a module-scope \`TextEncoder\` singleton (matching the existing pattern in \`src/universal/request.ts\`) and wrapped all 8 string enqueue calls with \`encoder.encode()\`. Binary paths through \`enqueueBinaryChunk\` are unchanged; they already produce \`Uint8Array\`.

## Tests

The existing stream tests read chunks via \`getReader()\` and compared with \`.toString()\`, which returns decimal byte notation for \`Uint8Array\` (e.g., \`"97"\` instead of \`"a"\`). Updated 6 tests with a \`decode\` helper that uses \`TextDecoder\` when the chunk is \`Uint8Array\`.

30/30 stream tests pass. Full suite: 1524 pass, 1 pre-existing fail in \`test/sucrose/integration.test.ts\` (spawns a Bun subprocess, exits 127 in WSL because the Bun binary on PATH is the Windows executable; unrelated to this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured consistent encoding of streamed textual outputs so all text-based stream formats (SSE, JSON, plain text) are transmitted reliably as binary-encoded chunks.

* **Tests**
  * Updated stream tests to normalize and assert decoded chunk contents, validating consistent behavior across different response types.

### DIFF
--- a/src/adapter/utils.ts
+++ b/src/adapter/utils.ts
@@ -213,6 +213,8 @@ interface CreateHandlerParameter {
 	mapCompactResponse(response: unknown, request?: Request): Response
 }
 
+const encoder = new TextEncoder()
+
 const enqueueBinaryChunk = (
 	controller: ReadableStreamDefaultController,
 	chunk: unknown
@@ -345,17 +347,17 @@ export const createStreamHandler =
 					// @ts-ignore
 					if (init.value.toSSE)
 						// @ts-ignore
-						controller.enqueue(init.value.toSSE())
+						controller.enqueue(encoder.encode(init.value.toSSE()))
 					else if (enqueueBinaryChunk(controller, init.value)) return
 					else if (typeof init.value === 'object')
 						try {
 							controller.enqueue(
-								format(JSON.stringify(init.value))
+								encoder.encode(format(JSON.stringify(init.value)))
 							)
 						} catch {
-							controller.enqueue(format(init.value.toString()))
+							controller.enqueue(encoder.encode(format(init.value.toString())))
 						}
-					else controller.enqueue(format(init.value.toString()))
+					else controller.enqueue(encoder.encode(format(init.value.toString())))
 				},
 
 				async pull(controller) {
@@ -384,17 +386,17 @@ export const createStreamHandler =
 						// @ts-ignore
 						if (chunk.toSSE)
 							// @ts-ignore
-							controller.enqueue(chunk.toSSE())
+							controller.enqueue(encoder.encode(chunk.toSSE()))
 						else if (enqueueBinaryChunk(controller, chunk)) return
 						else if (typeof chunk === 'object')
 							try {
 								controller.enqueue(
-									format(JSON.stringify(chunk))
+									encoder.encode(format(JSON.stringify(chunk)))
 								)
 							} catch {
-								controller.enqueue(format(chunk.toString()))
+								controller.enqueue(encoder.encode(format(chunk.toString())))
 							}
-						else controller.enqueue(format(chunk.toString()))
+						else controller.enqueue(encoder.encode(format(chunk.toString())))
 					} catch (error) {
 						console.warn(error)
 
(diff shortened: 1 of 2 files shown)


Zelys-DFKH:
🔔 Gentle ping — this PR has been ready with green CI for several days. Happy to address feedback or answer questions about the approach!

Issue #1771, opened 2026-03-01
v1.4.26 breaks Bun HTML imports


mpoc:
### What version of Elysia is running?

v1.4.26

### What platform is your computer?

Linux 6.6.87.2-microsoft-standard-WSL2 x86_64 x86_64

### What environment are you using

Bun v1.3.10

### Are you using dynamic mode?

No

### What steps can reproduce the bug?

Install Elysia v1.4.26, set up files:

index.html:

```html
<!doctype html>
<html>
  <body>
    <h1>Hello</h1>
  </body>
</html>
```

index.ts:
```typescript
import { Elysia } from "elysia";
import html from "./index.html";

new Elysia().get("/", html).listen(8000);
```
Run `bun run index.ts`, open http://localhost:8000/, page returns `{}`.

Same on v1.4.27.

Works as expected (page returns "Hello") on v1.4.25.

vermaysha:
@SaltyAom same with version 1.4.26

SaltyAom:
should be fixed in 1.4.28 (not published yet, please leave this PR open until 1.4.28 is published and you can verified locally)

mpoc:
Have verified this is fixed in v1.4.28. Thanks!

Teyik0:
Seems to be a regression on elysia 2 beta on this, got the same problem when trying to migrate

PR #1873, opened 2026-05-11
Release 2.0: Kiana

SaltyAom:
Release name: [DayDream](https://youtu.be/3qRd7Ok0NVY)

Code name: [Kiana](https://houkai2nd.miraheze.org/wiki/Kiana_Kaslana)

Status: Beta (waiting for blog post)

<a href="https://houkai2nd.miraheze.org/wiki/Kiana_Kaslana">
image
</a>

###### 45,873,632,010,000 until a miracle is made


SaltyAom:
# Elysia 2.0 "DayDream" *Early* Migration Guide

IMG_1967

## Recommendation
Use codemod CLI to automate most of the work

```ts
bunx @elysia/codemod@latest
```

That should do most of the work

There might be something that need manual work like new error handling syntax or macro stuffs

## 1. Hook and handler argument are swapped

The verb methods take the **schema/hook object before the handler**.

Applies to `get`/`post`/`put`/`patch`/`delete`/`options`/`head`/`all`/`method`

**Migration:** swap position of hook and handler

```ts
// 1.x
app.get('/user/:id', ({ params, query }) => params.id, {
  params: t.Object({ id: t.Number() }),
  query: t.Object({ name: t.String() })
})

// 2.0
app.get('/user/:id', {
  params: t.Object({ id: t.Number() }),
  query: t.Object({ name: t.String() })
}, ({ params, query }) => params.id)

// no hook, unchanged
app.get('/', () => 'hi')
```

---

## 2. WebSocket

- `websocket` is now opt-in plugin and need to apply at least once in any part of app
- `.ws()` now accept 2-3 arguments
	- 2-arg forms (`.ws('/ws', handler)`, `.ws('/ws', options)`) are unchanged
	- 3-arg form is now `.ws('/ws', options, handler)`
- `ws.data` is now inline to `ws` directly
- generator function and `yield` is now preferred way of sending data instead of `ws.send` for type safety

```ts
// 1.x
new Elysia()
	.ws('/', {
		message({ send, data: { params: { id } } }) {
			send(id)
		}
	})

// 2.x
import { websocket } from 'elysia/websocket'

new Elysia()
	.use(websocket())
	.ws('/', ({ params: { id } }) => id)
```

## 3. Drop `on` prefix for event

Use the bare-named methods.

| Removed | Use |
|---|---|
| `onRequest` | `request` |
| `onParse` | `parse` |
| `onTransform` | `transform` |
| `onBeforeHandle` | `beforeHandle` |
| `onAfterHandle` | `afterHandle` |
| `onAfterResponse` | `afterResponse` |
| `onError` | `error` |
| `onStart` | `setup` |
| `onStop` | `cleanup` |

```ts
// 1.x
app.
	onRequest(fn)
	.onBeforeHandle(fn)
	.onAfterResponse(fn)

// 2.0
app
	.request(fn)
	.beforeHandle(fn)
	.afterResponse(fn)
```

---

## 4. Code is drop from error

- `error(ErrorClass, fn)` registers a per-class handler
- `error(fn)` registers the general handler
- **`error.code` is removed**: dispatch with `instanceof` or `error(SomeError, fn)`

```ts
// 1.x
app
	.onError(({ code, error }) => {
	  	if (code === 'NOT_FOUND') return 'nope'
	})

// 2.0
app.error(NotFound, () => 'nope')
app.error(({ error }) => { /* ... */ })
```

---

## 5. `resolve` removed entirely

- `derive` now run on `beforeHandle` (previous `resolve` behavior)
- `resolve` are now removed

```ts
// 1.x
app.resolve(({ headers }) => ({ user: auth(headers) }))
// 2.0
app.derive(({ headers }) => ({ user: auth(headers) }))
```

## 6. Scope changes

- The `'scoped'` scope is **renamed to `'plugin'`**
- The `{ as: 'scope' }` **object form is removed**
- `.decorate()` / `.state()` `{ as: 'append' | 'override' }`

```ts
// 1.x
app.beforeHandle({ as: 'scoped' }, fn)
app.as('scoped')
app.guard({ as: 'scoped' }, fn)
app.decorate({ as: 'override' }, 'db', db)

// 2.0
app.beforeHandle('plugin', fn)
app.as('plugin')
app.guard('plugin', fn)
app.decorate('override', 'db', db)
```

---

## 7. Guard/group default to OVERRIDE channel

**Every** `.guard()` and `.group()` is default to **override**

- the closer to the route, the more power, the nearer schema **replaces** an inherited one
- `schema: 'standalone'` is explicitly required across all APIs

**Migration:** if you use `guard`, add `schema: 'standalone'`:

```ts
// 1.x string-scope & run forms were implicitly additive (standalone)
app.guard({ body: t.Object({ a: t.String() }) })

// 2.0 now overrides by default; prefer closer schema first
app.guard({ schema: 'standalone', body: t.Object({ a: t.String() }) })
```

---

## 8. TypeBox 1.0 alignment

See [TypeBox 1.0 migration](https://github.com/sinclairzx81/typebox/blob/main/changelog/1.0.0-migration.md)

But to summarize:
- `t.Transform` → **`t.Codec`**.
- Removed `t.Recursive`, `t.Not`, `t.RegExp`
- **`t.NoValidate` semantics changed:** now skips `Check` only `Default`/`Convert`/`Decode`/`Encode` still run.
	- `NoValidate` will throw `ValidationError` if Encode runs.
- `Error.summary` now uses TypeBox's default message (and supports Standard Schema).

---

## 9. Macros

- **Functional macro must use the object form:** `.macro({ name: fn })`.
- **`.macro(name, definition)` removed**, use `.macro({ [name]: definition })`.
	- The object form is now fully inferred (own schema, derive results, function-form arg, typo rejection)
- **Requires TypeScript ≥ 5.7** for macro definition inference.

```ts
// 1.x named form
app.macro('auth', {
  resolve: ({ headers }) => ({ user: auth(headers) })
})

// 2.0 object form only (fully inferred)
app.macro({
  auth: {
    resolve: ({ headers }) => ({ user: auth(headers) })
  }
})
```

## 10. Removed / renamed APIs & exports

- `NotFoundError` → **`NotFound`** (a thrown 404 is `instanceof NotFound`)
- `getSchemaValidator` → **`Validator.create`**
- `set.redirect`: use the `redirect()` context helper
- `context.contentType` is removed from `Context` in `parse`, use `context.contentType` instead
- Deprecated `response` field on `mapResponse`/`afterResponse` is removed
- Removed instance methods `.route()`, `.connect()`, `.env()`,
  `.affix()`/`.prefix()`/`.suffix()` and the `.store`/`.decorator`/`.config`
  **instance getters**, use the verb methods, `new Elysia({ name, prefix })` for
  plugin naming, and in-handler `context.store`/`context.decorator`
- Passing an Elysia instance to `.mount` is deprecated, use `.use` instead
- `config.encodeSchema` dropped (always enabled now)

```ts
import { NotFound } from 'elysia'          // was NotFoundError
import { Validator } from 'elysia'         // getSchemaValidator → Validator.create

// redirect
;({ set }) => { set.redirect = '/' }       // 1.x
;({ redirect }) => redirect('/')           // 2.0

// afterResponse / mapResponse field
app.afterResponse(({ responseValue }) => {})   // was `response`

// parse contentType
app.parse((ctx, contentType) => {})        // 1.x (2nd param)
app.parse((ctx) => ctx.contentType)        // 2.0

// removed instance getters / methods → use context or constructor
new Elysia({ name: 'x', prefix: '/v1' })   // was .prefix()/.affix()/.suffix()/name
;({ store, decorator }) => {}              // was app.store / app.decorator getters

// mount an instance
app.use(plugin)                            // was app.mount(plugin)
```

---

## 11. Behavior changes worth knowing (no API change, but observable)

- **`afterHandle` skips the rest on short-circuit.**
- **Bodyless `GET`/`HEAD` no longer run `parse` hooks**, a `parse` hook that must run on every request belongs in `request`/`transform`
- **422 response no longer echoes large bodies**, `found` is scoped; a request body whose JSON exceeds 4KB isn't reflected back in full. `error.value` still exposes the full value to custom handlers.
- **Returned `Response` passed by reference when `set` is untouched** (preserves
  `content-length`).
	- **Trap:** returning the *same* `Response` object across requests while writing per-request `set.headers` now mutates the shared object and leaks headers (previously errored loudly). Return a fresh `Response` per
  request.
- **`streamResponse` yields raw bytes**, re-streamed bodies are now byte-identical; a headerless binary-first chunked stream defaults to `application/octet-stream` (was `application/json`)
- **WS query parsing** now matches HTTP (duplicate keys → arrays, not last-wins; null-prototype records)
- **Validator runs `Convert → Check → DecodeUnsafe`** for codec schemas (fixes `t.Numeric({ minimum, maximum })` and `t.Codec(...).Decode(...)` against wire input)
- `file-type` is no longer shipped by default. If `t.File({ type: 'string' })` is used without file type detector, Elysia will warn to set `setFileTypeDetector()` which can run by adding `fileTypeFromBlob` from `file-type` package

agusgarcia3007:
This is hugeeeeee. so anxious to have a stable version to upgrade Learnbase's server. LFGG

rubenferreira97:
Really liking the direction of 2.0 overall and most of these changes feel like improvements. The only one I'm not sold on is dropping the `on` prefix for lifecycle events.

| Removed | Alternative |
|----------|----------|
| `onRequest` | `onRequest` |
| `onParse` | `onParse` |
| `onTransform` | `onTransform` |
| `onBeforeHandle` | `beforeHandle` |
| `onAfterHandle` | `afterHandle` |
| `onAfterResponse` | `afterResponse` |
| `onError` | `onError` |

For me, `onRequest`, `onParse`, and `onError` immediately communicate that these are event hooks, while `request`, `parse`, and `error` feel more like actions. Sometimes saving a few characters isn't worth it if it comes at the cost of readability and discoverability.

Necrelox:
> Really liking the direction of 2.0 overall and most of these changes feel like improvements. The only one I'm not sold on is dropping the `on` prefix for lifecycle events.
> 
> Removed	Alternative
> `onRequest`	`onRequest`
> `onParse`	`onParse`
> `onTransform`	`onTransform`
> `onBeforeHandle`	`beforeHandle`
> `onAfterHandle`	`afterHandle`
> `onAfterResponse`	`afterResponse`
> `onError`	`onError`
> For me, `onRequest`, `onParse`, and `onError` immediately communicate that these are event hooks, while `request`, `parse`, and `error` feel more like actions. Sometimes saving a few characters isn't worth it if it comes at the cost of readability and discoverability.

I think `beforeHandle`, `afterHandle`, and `afterResponse` should also be prefixed with `on`

Either way, this new version looks really promising, and I can't wait to use it !!

rubenferreira97:
> I think beforeHandle, afterHandle, and afterResponse should also be prefixed with on
> 
> Either way, this new version looks really promising, and I can't wait to use it !!

I’d be fine with that as well. Both options solve the "verb/action" issue and come with a few advantages:
- When typing `on*`, it lists all events.
- No breaking changes.

SaltyAom:
> Really liking the direction of 2.0 overall and most of these changes feel like improvements. The only one I'm not sold on is dropping the `on` prefix for lifecycle events.
> 
> Removed	Alternative
> `onRequest`	`onRequest`
> `onParse`	`onParse`
> `onTransform`	`onTransform`
> `onBeforeHandle`	`beforeHandle`
> `onAfterHandle`	`afterHandle`
> `onAfterResponse`	`afterResponse`
> `onError`	`onError`
> For me, `onRequest`, `onParse`, and `onError` immediately communicate that these are event hooks, while `request`, `parse`, and `error` feel more like actions. Sometimes saving a few characters isn't worth it if it comes at the cost of readability and discoverability.

We are doing this to align all events eg. `mapResponse` (is v1 not `onMapResponse`), `derive` not `onDerive`. So to align them, we have to choose either to add `on` or drop `on`

Additionally, we want to align the route hook/guard and APIs:
```ts
new Elysia()
    // match route hook
    .transform(() => {})
    .get('/', {
        transform() {}
    }, () => 'ok)
```

We decided that most of the events sound better without `on` and if we rename everything to `on`, we also have to rename the route hook as well as we are trying to align the naming

Most of the events don't really need `on` and still communicate that it will run on the event, eg. `mapDerive`, `afterResponse`, `beforeHandle`, `transform`, `trace` which are not aligned in v1

`error` are understable as we introduce a new API
```ts
new Elysia()
    .error(NotFound, () => ':(')
```

Only concern I have is `request` that can sound ambiguous

3160C240-58F9-45F7-BDDB-D117625F949C_4_5005_c

Necrelox:
Ah, that makes more sense now, i understand the idea better, but I agree that `request` feels a bit ambiguous.
Maybe something like `incomingRequest` (or just `incoming`) or `receive` would be more explicit 😕

rubenferreira97:
> We are doing this to align all events eg. `mapResponse` (is v1 not `onMapResponse`), `derive` not `onDerive`. So to align them, we have to choose either to add `on` or drop `on`.

I agree that the current API is mixed and that improving consistency for 2.0 is a good goal. But I do not think dropping `on` makes the API more consistent overall, because it loses another kind of consistency: the distinction between actions and lifecycle handler registration.

`onRequest`, `onParse`, and `onError` immediately read as lifecycle hooks: this code runs when that event happens.

```ts
app.onRequest(fn)
app.onParse(fn)
app.onError(fn)
```

Names like `request`, `parse`, `transform`, and `error` read more like actions. The app is not requesting, parsing, transforming, or erroring at the call site. It is registering handlers for those lifecycle points.

```ts
app.request(fn)
app.parse(fn)
app.transform(fn)
app.error(fn)
```

I agree that aligning app hooks with route hooks is valuable:

```ts
new Elysia()
    .transform(() => {})
    .get('/', {
        transform() {}
    }, () => 'ok')
```

But route-local hooks already have context because they live inside the route options object. Chainable methods like `.request(fn)` or `.error(fn)` have less context, so the `on` prefix helps readability and discoverability.

`mapResponse`, `derive`, and `mapDerive` also feel different to me. They read more like behavior-oriented APIs: map this response, derive this context. `request`, `parse`, and `error` are more ambiguous as bare chainable methods.

For example, `error(NotFound, fn)` is a nice API:

```ts
new Elysia()
    .error(NotFound, () => ':(')
```

but `.error(...)` is still less obvious than `.onError(...)` unless you already know the convention. The same concern is even stronger for `request`, which can sound ambiguous.

So my preference would be to either:

1. Keep `on*` for lifecycle event registration.
2. Keep `on*` aliases as supported APIs.
3. Or find clearer names for the ambiguous cases. (As a programmer I know naming is one of the hardest parts here 🙂)

Based on the upvotes 👍 and follow-up agreement, I think this is worth discussing before 2.0 stabilizes.

## What other frameworks are doing

<details>
<summary>Fastify</summary>

Their hooks are not all `on*`, but they are still understandable because the naming follows context:

```ts
fastify.addHook('onRequest', fn)
fastify.addHook('preParsing', fn)
fastify.addHook('preValidation', fn)
fastify.addHook('preHandler', fn)
fastify.addHook('onError', fn)
fastify.addHook('onSend', fn)
fastify.addHook('onResponse', fn)
fastify.addHook('onReady', fn)
fastify.addHook('onClose', fn)
fastify.addHook('onRoute', fn)
```

So I do not think `on` has to be all-or-nothing. Context matters. `preHandler` makes sense because it describes a phase before handling. `onRequest` and `onError` make sense because they describe reacting to an event.

</details>

<details>
<summary>Nitro</summary>

Nitro is probably the closest counterexample because it uses bare event names:

```ts
nitroApp.hooks.hook('request', fn)
nitroApp.hooks.hook('response', fn)
nitroApp.hooks.hook('error', fn)
nitroApp.hooks.hook('close', fn)
```

But Nitro gets away with this because the registration context is explicit:

```ts
hooks.hook('request', fn)
```

`request` is clearly the name of an event passed to a hook registrar. It does not read like the app itself is performing a request.

That is different from:

```ts
app.request(fn)
app.error(fn)
```

where the method name itself has to communicate whether it is an action or a lifecycle registration.

</details>

<details>
<summary>Koa</summary>

Koa uses middleware for request flow:

```ts
app.use(fn)
```

but event subscription uses `on`:

```ts
app.on('error', fn)
```

So `on` is preserved where the API is event-listener registration.

</details>

<details>
<summary>h3</summary>

h3 uses explicit `on*` global hooks:

```ts
onRequest
onError
onResponse
```

</details>

TL;DR: I could not find a framework that uses direct chainable lifecycle methods like `.request(fn)` or `.error(fn)`.

devSparkle:
Overall great changes!

However, I can't begin to stress how much more unreadable the code becomes with the removal of `on`.

Keeping it, adding it to methods that are missing it, or even going a step further and making the naming more explicit (like `hookRequest(fn)`) would make everything both more consistent, and easier to parse through.

mrReiha:
Thank you first of all.

But a quick question. what's the actual idea behind introducing a new `.error( NotFound, () => '404' )` and removing `code`? now what happens to the main `.error( ({ error }) => { ... } )`? if we have `NotFound` (per-class error) set, then those errors won't actually trigger the main error listener? or they're duplicated? then we should filter all of our previously set (per-class) errors in the main handler?

The thing is, I think IMHO apps usually have a main error listener to handle everything not handled anywhere else. so they need some sort of "thing" (it was `code` before v2) to be able to distinguish what's happening in the main handler. and if the app already handled some errors (with the newly introduced way of handling errors), then are we forced to filter all of those from main handler, again?

and if they're not duplicated (`NotFound` errors only triggers the specific handler for themselves) then there is another scenario, when maybe some route/plugins have their own dedicated `NotFound` and some others don't. So, main error listener still needs to count for `NotFound`s to be sure? in this case, some 404 errors hit the main listener and some don't?

## Summary:

If regardless of any other `.error( ErrorClass, fn )` set, main `.error( fn )` still gets triggered for EVERY error, then are we scattering error handling in all our app with this new introduced api? how are we supposed to know if a certain route/plugin is already handled some specific error or not? or it just doesn't matter and I'm overseeing it?

I didn't quite understand and just wanted to bring the subject up, maybe someone else had the same question <3

SaltyAom:
> Thank you first of all.
> 
> But a quick question. what's the actual idea behind introducing a new `.error( NotFound, () => '404' )` and removing `code`? now what happens to the main `.error( ({ error }) => { ... } )`? if we have `NotFound` (per-class error) set, then those errors won't actually trigger the main error listener? or they're duplicated? then we should filter all of our previously set (per-class) errors in the main handler?
> 
> The thing is, I think IMHO apps usually have a main error listener to handle everything not handled anywhere else. so they need some sort of "thing" (it was `code` before v2) to be able to distinguish what's happening in the main handler. and if the app already handled some errors (with the newly introduced way of handling errors), then are we forced to filter all of those from main handler, again?
> 
> and if they're not duplicated (`NotFound` errors only triggers the specific handler for themselves) then there is another scenario, when maybe some route/plugins have their own dedicated `NotFound` and some others don't. So, main error listener still needs to count for `NotFound`s to be sure? in this case, some 404 errors hit the main listener and some don't?
> 
> ## Summary:
> If regardless of any other `.error( ErrorClass, fn )` set, main `.error( fn )` still gets triggered for EVERY error, then are we scattering error handling in all our app with this new introduced api? how are we supposed to know if a certain route/plugin is already handled some specific error or not? or it just doesn't matter and I'm overseeing it?
> 
> I didn't quite understand and just wanted to bring the subject up, maybe someone else had the same question <3

> what's the actual idea behind introducing a new `.error( NotFound, () => '404' )` and removing `code`? 

type safety

`code` is not deterministic in type-level and the new syntax fix that

For example, if you return custom error from handler, the new syntax will coerce it to Eden response which is what v1 can't do (see attached photo for reference)

image

> if we have `NotFound` (per-class error) set, then those errors won't actually trigger the main error listener?

Same mechanism as v1, order by registration

> if the app already handled some errors (with the newly introduced way of handling errors), then are we forced to filter all of those from main handler, again?

No, similar to v1. If you have multiple error handler functions, the function will loop through until some function return something and use it as response

Necrelox:
Hello @SaltyAom, regarding https://github.com/elysiajs/elysia/pull/1557, I was wondering if it would be possible in v2 to expose route metadata through macro introspection especially the resolved route path and HTTP method

The goal would be for a macro to know which route it is being applied to 🥺

SaltyAom:
Screenshot 2569-07-09 at 08 21 55
Screenshot 2569-07-09 at 08 43 54
Screenshot 2569-07-09 at 08 45 39
Screenshot 2569-07-09 at 08 47 09

SaltyAom:
#1941 one more patch bro, please just one more patch

Necrelox:
Maybe removing the precompile option from Elysia would make sense now that there's the AOT plugin?

rubenferreira97:
@SaltyAom I noticed a possible inconsistency around the optional TypeBox peer dependency in Elysia 2.

I am using `elysia@2.0.0-exp.61` with Valibot through Standard Schema, under Bun's isolated linker. TypeBox is not used directly by the application.

The package declares `typebox` as an optional peer dependency, but importing only `Elysia` from the main entrypoint still loads TypeBox modules:

```ts

new Elysia()
```

Without `typebox` installed, this fails at runtime with an error similar to:

```text
Cannot find module 'typebox/type' from
'elysia/dist/type/validator/index.mjs'
```

The main `elysia` entrypoint appears to import and re-export `t`, `TypeSystem`, and `TypeBoxValidator` eagerly, so TypeBox is loaded even when none of those exports are used.

Using the core subpaths works without TypeBox:

```ts
import { Elysia } from 'elysia/base'
import { status } from 'elysia/error'

const app = new Elysia().get('/', () => status(200, 'ok'))
```

I also noticed that `@elysia/cors` already imports `Elysia` from `elysia/base`.

Is the main `elysia` entrypoint intentionally TypeBox-dependent, with `elysia/base` being the recommended entrypoint for applications using only Standard Schema validators such as Valibot or Zod?

If so, it might be useful to document this distinction. Otherwise, would it be possible for the main entrypoint to avoid loading TypeBox unless the TypeBox-specific exports are used?

Thanks!

codler:
Great work with 2.0!

Im using `elysia@2.0.0-beta.2` and `@elysia/eden@2.0.0-beta.2` and im getting 
```
npm error Could not resolve dependency:
npm error peer elysia@">= 2.0.0-exp.23" from @elysia/eden@2.0.0-beta.2

**Comment by codler** [comment]
From https://github.com/elysiajs/elysia/pull/1873#issuecomment-4734573873
> ## 10. Removed / renamed APIs & exports
> ```ts
> // mount an instance
> app.use(plugin)                            // was app.mount(plugin)
> ```
> 

When I rename mount to use
```ts
export const authMiddleware = new Elysia({ name: "better-auth" })
  .use(auth.handler)
```
im start getting 404 on my routes and this message
```
TypeError: Invalid URL
    at ignore-listed frames {
  code: 'ERR_INVALID_URL',
  input: 'undefined'
}
```
I use 2.0.0-beta.4

raghu2x:
Unable to start dev server with static Plugin. I am using elysia@2.0.0-beta.4

image

image

raghu2x:
Route context parameter is loosing it's type when i use authMiddleware. When i remove the bearer plugin from authMiddleware it worked. I think this issue is with bearer plugin.

image

image

raghu2x:
Open api documentation not generated for even a single api, With `elysia@1.4.29` it was working fine. in `elysia@2.0.0-beta.4` it's not working.

image

SaltyAom:
@raghu2x static plugin is now fixed. I'm unable to reproduce problem with bearer and openapi. Might need a PoC

raghu2x:
@SaltyAom  Elysia codemod is not working with bun. with bunx it migrates package.json file only and no error or warning nothing. 

`bunx @elysia/codemod@latest`  -  It's not working 

`npx @elysia/codemod@latest`   - It's working 

My specs
Platform - Windows 11
Bun - v1.3.14
Node - v24.15.0
IDE Terminal - VS Code - Powershell
Elysia - v1.4.29

image

raghu2x:
@SaltyAom  Here is the reproduction of 3 issues

```typescript
import { bearer } from '@elysia/bearer';
import cors from '@elysia/cors';
import { openapi } from '@elysia/openapi';
import { Elysia, t } from 'elysia';

const userRoutes = new Elysia({
  //ISSUE 1: If we assign tag here we were getting at per route level in elysia v1
  // but in v2 this have no effect we need to define tags on each route
  tags: ['Users']
})
  .use(bearer())

  .get(
    '/users/:id',
    {
      params: t.Object({
        id: t.Number({ description: 'Numeric id of the user to fetch', examples: [1] })
      }),
      detail: {
        // tags: ['Users'],
        summary: 'Get a user by id'
      }
    },
    // ISSUE 2: When we used bearer plugin params lost there type
    ({ params }) => ({ id: params.id, name: `User ${params.id}` })
  );

// ISSUE 3: When we use a separate elysia instance for plugin registration openapi not working
// Directly using on app openapi works & in v1 it was working with separate instance
const configElysia = new Elysia().use(cors()).use(openapi());

const app = new Elysia({})
  .use(configElysia)

  .get('/', () => 'Hello Elysia')
  .use(userRoutes)

  .listen(3000);

console.log(`🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`);

```

codler:
Love the new TreatyPlugin in Eden 2.0, now i can do like api.users.get.queryOptions()

Teyik0:
@SaltyAom seems to be a regression about this on Elysia 2 => https://github.com/elysiajs/elysia/issues/1771
The issue is close so I'm not sure you'll see it, though I'm posting here the comment

Necrelox:
Hello @SaltyAom ,

I have a small proposal to Elysia V2 🥺🥺🥺 regarding schemas, guards, and macro

It would be useful to support a union of `t.Ref(...)` references to multiple models registered with `.model({...})`, for example:

```ts
response: {
  401: t.Union([
    t.Ref('UnauthorizedResponse'),
    t.Ref('BusinessErrorResponse')
  ])
} // actually is not work 😟
```

It would also be useful to add a composition mode for guards:

```ts
schema: "override"; // default and current behavior
schema: "merge"; // merge/intersect (allOf behavor)
schema: "union"; // A | B (anyOf behavior)
```

with `schema: 'union'`, if both a guard and a route define a response for the same status code, the generated OpenAPI specification would use `anyOf` rather than `allOf`

Ideally, this behavior could apply to `body`, `params`, `query`, `headers`, `cookie`, and `response`, and also be available in macros 🥺 this would allow a macro to add a `401` response without preventing a route from adding its own alternatives

Necrelox:
Following up on my small V2 proposal https://github.com/elysiajs/elysia/pull/1873#issuecomment-5409108279), i also noticed something related to the OpenAPI documentation, more specifically with some validation methods such as `t.Nullable()`, which internally use `elyType`

In `src/type/elysia/utils.ts` (l. 33):

```ts
export function elyType<T extends TSchema>(
	name: ELYSIA_TYPES[keyof ELYSIA_TYPES],
	schema: T
): T {
	const target = Object.assign(
		Object.create(Object.getPrototypeOf(schema)),
		schema,
		{ '~elyTyp': name }
	) as T

	copyNonEnumerable(schema, target, '~elyTyp')

	return target
}
```

Here, `~elyTyp` is added directly through `Object.assign`:

```ts
{ '~elyTyp': name }
```

This makes the property **enumerable**, which can cause it to appear in the generated OpenAPI spec
Would it make more sense to keep this property as non-enumerable internal metadata instead ?

For example:

```ts
export function elyType<T extends TSchema>(
	name: ELYSIA_TYPES[keyof ELYSIA_TYPES],
	schema: T
): T {
	const target = Object.assign(
		Object.create(Object.getPrototypeOf(schema)),
		schema
	) as T

	copyNonEnumerable(schema, target, '~elyTyp')

	Object.defineProperty(target, '~elyTyp', {
		value: name,
		enumerable: false,
		writable: true,
		configurable: true
	})

	return target
}
```

This way, `~elyTyp` would still be available internally, while avoiding exposing it during schema enumeration and, as a result, in the generated OpenAPI spec

Necrelox:
https://github.com/elysiajs/elysia/pull/1873#issuecomment-5440329394
https://github.com/elysiajs/elysia/pull/1873#issuecomment-5409108279

Hello @SaltyAom ! Following up on these two messages 🥺

Now that the `QUERY` method is finally available, would it be possible to add support for it in Elysia please ?
It would make it possible to define routes with more complex filters without having to fall back to `POST`.
It would also avoid having to add an artificial suffix like `/search` just because there would otherwise be two `POST` routes on `/`: one for searching and one for inserting 😅

SaltyAom:
@Necrelox QUERY method has been added on latest beta. As for union schema. The problem is that the union schema has a lot of potential footgun, as it could easily multiply all possible type workload very quickly and would cause the "typescript excessively deep and possibly infinite type" error

Necrelox:
@SaltyAom Niceeee, that's great for QUERY !! I'll be able to implement it, and I'll wait for the OpenAPI plugin to support 3.2 before using it in production ! ❤️

I see what you mean about unions. It's a bit unfortunate since they would've been pretty useful, but I understand the constraints

Also, is there a way to make a union of refs like this?

```ts
response: {
  401: t.Union([
    t.Ref('UnauthorizedResponse'),
    t.Ref('BusinessErrorResponse')
  ])
} // doesn't currently work 😟
```

Right now I have to duplicate the schemas as a workaround 😬

Issue #313, opened 2023-11-15
Option to return errors as JSON


hassanshaikley:
Hey, it would be nice to be able to do that.
I understand the onError callback but that requires a good bit of finessing to convert to JSON.
In particular I am thinking of validator errors.

SaltyAom:
I'm not sure if you mean or:
1. Returning Error from handler as JSON
2. Returning built-in Elysia errors like ValidationError, NotFound as JSON.

Either way, you should be able to use https://elysiajs.com/patterns/error-handling.html#catching-all-error

oSethoum:
what he means is for elysia to automatically send back the validation error as JSON not in plain text
what should I do with this in my client side app? parse it?
![image](https://github.com/elysiajs/elysia/assets/88779394/ae5ac08a-89ad-4812-a6fb-23a03c302fab)

hassanshaikley:
Exactly @oSethoum! 

If not automatically then it should be easier. I suppose I can look into creating a plugin for this?

wiredmatt:
import { Elysia, t } from "elysia";
import { PrismaClient } from "@prisma/client";

const db = new PrismaClient();

const app = new Elysia()
  .post(
    "/sign-up",
    async ({ body }) =>
      db.user.create({
        data: body,
      }),
    {
      body: t.Object({
        username: t.String(),
        password: t.String({
          minLength: 8,
        }),
      }),
      error({ code, error }) {
        switch (code) {
          case "VALIDATION":
            console.log(error.all);

            // Find a specific error name (path is OpenAPI Schema compliance)
            const name = error.all.find((x) => x.path === "/name");

            // If has validation error, then log it
            if (name) console.log(name);

            return error.all;
          default:
            return {
              name: error.name,
              message: error.message,
            };
        }
      },
    }
  )
  .listen(3000);

console.log(
  `🦊 Elysia is running at ${app.server?.hostname}:${app.server?.port}`
);
![image](https://github.com/elysiajs/elysia/assets/29718978/89bff503-7d54-4922-9da3-5a014266b8c3)

EvHaus:
I'm also looking for this support. At the moment, I can hack around it by adding this at the top of the Elysian chain:

```ts
.onError(({ error }) => {
	return Response.json({error: error.message});
})
```

But this doesn't get handled nicely by Eden since it doesn't know that `{error: string}` will now be a valid response type from all API endpoints, so you end up needing to hack around the missing types on the client:

```ts
[...]

PR #1887, opened 2026-05-25
fix: type plain onError returns as error responses

artrmkr:
/claim #313
@algora-pbc /claim #313

## Summary
- infer plain values returned from `onError` as default `500` error responses instead of successful `200` responses
- keep explicit `status(...)` values from error handlers under their declared status codes
- apply the same error-handler response inference to route-level `error` hooks in guards/groups

## Tests
- `npx bun run test:types`
- `npx prettier --check src/index.ts src/types.ts test/types/lifecycle/soundness.ts`
- `git diff --check`

Fixes #313

## Summary by CodeRabbit

* **Refactor**
  * Improved public TypeScript typings for error-handler responses: new exported utilities now infer error return shapes and status mappings more precisely across lifecycle, group, and guard scenarios.

* **Tests**
  * Updated type-level tests to validate error-handler response inference, including correct 500 mappings, unions of explicit status/plain returns, and macro error return propagation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elysiajs/elysia/pull/1887?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -147,8 +147,11 @@ import type {
 	StandardSchemaV1Like,
 	ElysiaHandlerToResponseSchema,
 	ElysiaHandlerToResponseSchemas,
+	ElysiaErrorHandlerToResponseSchema,
+	ElysiaErrorHandlerToResponseSchemas,
 	ExtractErrorFromHandle,
 	ElysiaHandlerToResponseSchemaAmbiguous,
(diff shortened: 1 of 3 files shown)


artrmkr:
@algora-pbc /claim #313

PR #1914, opened 2026-06-08
fix: map onError plain returns to error status codes instead of 200

ClankOS:
## Summary

Fixes #313

- When `onError` returns a plain value (not wrapped in `status()`), the type system previously recorded it under status `200`. At runtime, Elysia correctly preserves the error's actual status code via `if(set.status===200||!set.status) set.status=error.status`. This type/runtime mismatch caused Eden to type errors as `{status: unknown, value: unknown}` instead of the actual return type.
- Introduces `ErrorValueToResponseSchema` (and corresponding `ErrorHandlerToResponseSchema`, `ErrorHandlerToResponseSchemas`, `ErrorHandlerToResponseSchemaAmbiguous`) that maps plain `onError` returns to error status codes (400, 404, 422, 500) instead of 200
- Explicit `status()` returns (e.g., `status(404, 'not found')`) continue to work exactly as before
- Updates type tests to reflect the corrected behavior

## Example

```ts
const app = new Elysia()
  .onError(({ error }) => {
    return { failure: error.message }
  })
  .get('/', () => { throw new Error('oops') })
```

**Before:** Eden types `error` as `{status: unknown, value: unknown}` because `{failure: string}` was recorded under status 200.

**After:** `{failure: string}` is recorded under error statuses (400, 404, 422, 500), so Eden can properly type the error response.

## Test plan

- [x] `tsc --noEmit --project tsconfig.test.json` passes with zero errors
- [x] All 183 lifecycle tests pass (`bun test test/lifecycle/`)
- [x] All 213 core tests pass (`bun test test/core/`)
- [x] Updated type soundness tests for onError local/scoped/global scopes

@algora-pbc /claim #313

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Refactor**
  * Improved TypeScript inference for error handlers so plain error returns are reflected as standard error status responses (e.g., 400/422/500) and aggregated correctly across scoped/global handlers.
  * Added utilities to model combined/ambiguous error response shapes for more accurate editor hints.

* **Tests**
  * Updated type-level tests to align expectations with the refined error-response inference.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2192,6 +2192,26 @@ export type ValueToResponseSchema<Value> = ExtractErrorFromHandle<Value> &
 				: { 200: R200 }
 		: {})
 
+/**
+ * Like ValueToResponseSchema but for onError handlers.
+ * Maps plain (non-ElysiaCustomStatusResponse) returns to error status codes
+ * instead of 200, matching runtime behavior where error status is preserved.
+ */
+type ExtractPlainErrorReturn<T> = Exclude<
+	T extends AnyElysiaCustomStatusResponse
+		? Exclude<T, AnyElysiaCustomStatusResponse>
+		: T,
+	undefined
+>
+
+export type ErrorValueToResponseSchema<Value> =
+	ExtractErrorFromHandle<Value> &
+		(ExtractPlainErrorReturn<Value> extends infer ErrVal
+			? IsNever<ErrVal> extends true
+				? {}
+				: { 400: ErrVal; 404: ErrVal; 422: ErrVal; 500: ErrVal }
+			: {})
+
 export type ValueOrFunctionToResponseSchema<T> = T extends (
 	...a: any
 ) => MaybePromise<infer R>
@@ -2202,6 +2222,13 @@ export type ElysiaHandlerToResponseSchema<in out Handle extends Function> =
 	Prettify<
 		Handle extends (...a: any) => MaybePromise<infer R>
 			? ValueToResponseSchema<Exclude<R, undefined>>
+			: {}
+	>
+
+export type ErrorHandlerToResponseSchema<in out Handle extends Function> =
+	Prettify<
+		Handle extends (...a: any) => MaybePromise<infer R>
+			? ErrorValueToResponseSchema<Exclude<R, undefined>>
 			: {}
 	>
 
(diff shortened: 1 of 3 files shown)


cmiranda12 (approved):
I have a complete picture. Let me now write the consolidated review.

---

DECISION: approved_with_comments

## AI Code Review — elysia (full tier)

This is a focused, type-only fix that correctly addresses a clear bug: plain (non-`ElysiaCustomStatusResponse`) returns from `onError` handlers were previously mapped to status `200` in the inferred response schema, when the runtime actually preserves the original error's status code. The fix introduces a parallel set of type utilities (`ErrorHandlerToResponseSchema`, `ErrorHandlerToResponseSchemas`, `ErrorHandlerToResponseSchemaAmbiguous`, and their backing `ErrorValueToResponseSchema` / `ExtractPlainErrorReturn` helpers) that map plain returns to a set of error status codes instead of 200. All three lifecycle scopes (local/volatile, scoped/ephemeral, global/metadata) are exercised in the updated type tests, and the test expectations are mathematically consistent with the new type logic. No runtime code is changed.

### Suggestions

- **`src/types.ts:2213`** — Hard-coded error status set is a narrow approximation. `{ 400: ErrVal; 404: ErrVal; 422: ErrVal; 500: ErrVal }` covers the four most common Elysia built-in error codes, but excludes `401`, `403`, `405`, `429`, `502`, `503`, etc. If a plain return from an `onError` handler is associated with one of those codes at runtime, the type schema will silently miss it. This is inherent to the problem (the actual error code isn't statically knowable), but the limitation should be documented in the JSDoc on `ErrorValueToResponseSchema`. A brief note like *"Plain returns are typed against the most common Elysia error codes; explicit `status(N, value)` returns will always be captured accurately for any N"* would help future maintainers understand the trade-off rather than wondering whether the list is exhaustive.

- **`src/types.ts:2247`** — `ErrorHandlerToResponseSchemas` is a structural mirror of `ElysiaHandlerToResponseSchemas`, differing only in delegating to `ErrorHandlerToResponseSchema` instead of `ElysiaHandlerToResponseSchema`. Both retain the same `// @ts-ignore Trust me bro` suppression pattern. A short comment stating why a shared generic recursive type wasn't used (TypeScript variance / `in out` constraints on the handler parameter) would reduce future duplication questions.

_Generated by Froholdt Pi-based reviewer._

src/types.ts
```diff
@@ -2192,6 +2192,36 @@ export type ValueToResponseSchema<Value> = ExtractErrorFromHandle<Value> &
...
+ * etc.) are not statically knowable from the return type alone and will not appear
+ * in the schema. Use an explicit `status(N, value)` return to capture any specific
+ * status code accurately.
+ */
+export type ErrorValueToResponseSchema<Value> =
+	ExtractErrorFromHandle<Value> &
+		(ExtractPlainErrorReturn<Value> extends infer ErrVal
+			? IsNever<ErrVal> extends true
+				? {}
+				: { 400: ErrVal; 404: ErrVal; 422: ErrVal; 500: ErrVal }
```
SaltyAom, line 2222: slop

PR #1924, opened 2026-06-16
fix: validate file contents in dynamic mode

### LINKED ISSUE #1922: t.File/t.Files content validation is skipped in dynamic mode (aot: false)
## What version of Elysia is running?

`elysia@1.4.28`

Also reproduced against commit:

```txt
56310be9
```

## What platform is your computer?

```txt
Linux 7.0.12-201.fc44.x86_64 x86_64
```

## What environment are you using?

```txt
Bun: 1.3.9
file-type: 20.5.0
```

## Are you using dynamic mode?

Yes.

This issue reproduces with:

```ts
new Elysia({ aot: false })
```

The same schema behaves differently in AOT mode and dynamic mode.

AOT mode rejects the spoofed upload as expected. Dynamic mode accepts it.

## What steps can reproduce the bug?

`t.File({ type })` validates uploaded file contents using magic bytes in AOT mode, but in dynamic mode the same schema accepts a file whose bytes do not match the declared type.

Minimal reproduction:

```ts
import { Elysia, t } from 'elysia'

const aot = process.env.AOT !== 'false'

new Elysia({ aot })
  .post('/avatar', async ({ body }) => {
    const f = body.file as File

    return {
      ok: true,
      storedType: f.type,
      size: (await f.arrayBuffer()).byteLength
    }
  }, {
    body: t.Object({
      file: t.File({ type: 'image/png' })
    })
  })
  .listen(3000)
```

Create a spoofed file. The file contents are SVG, but the upload declares `image/png`:

```bash
printf '<svg xmlns="http://www.w3.org/2000/svg"><script>alert(document.domain)</script></svg>' > evil.png
```

Send it as multipart with a forged MIME type:

```bash
curl -i -F 'file=@evil.png;type=image/png;filename=avatar.png' http://localhost:3000/avatar
```

Run in dynamic mode:

```bash
AOT=false bun run index.ts
```

Run in AOT mode:

```bash
bun run index.ts
```

## What is the expected behavior?

Both execution modes should reject the upload with `422`.

The schema is the same:

```ts
t.File({ type: 'image/png' })
```

So the validation guarantee should be the same in both modes.

Expected result for the spoofed SVG uploaded as `image/png`:

```txt
HTTP 422
Invalid file type
```

## What do you see instead?

In AOT mode, the file is rejected:

```txt
HTTP 422
{
  "type": "validation",
  "on": "body",
  "summary": "Invalid file type",
  "message": "\"body.file\" has invalid file type",
  "property": "body.file",
  "expected": "image/png"
}
```

In dynamic mode, the same file is accepted:

```txt
HTTP 200
{"ok":true,"storedType":"image/png","size":85}
```

A valid 1x1 PNG was also tested in AOT mode and returned `200`, so AOT is not rejecting all uploads blindly. It correctly distinguishes a real PNG from a forged one.

## Additional information

This looks like a behavior mismatch between the AOT and dynamic validation pipelines.

From the codebase, `t.File({ type })` loads `file-type`, and the strong validation path is `fileType()`, which uses `fileTypeFromBlob(file)` to inspect the real file bytes.

In AOT mode, route composition emits a call equivalent to:

```ts
await fileType(c.body.file, "image/png", 'body.file')
```

In dynamic mode, body validation goes through TypeBox `Check()`:

```ts
if (validator.createBody?.()?.Check(body) === false)
  throw new ValidationError('body', validator.body!, body)
```

For `File`, that synchronous validator checks `value.type`, which is the MIME type provided by the client in the multipart request.

I could not find a `fileType()` call in `src/dynamic-handle.ts`:

```bash
grep -n "fileType" src/dynamic-handle.ts
# no result
```

So in dynamic mode, `t.File({ type: 'image/png' })` appears to collapse from magic-byte validation to checking the client-controlled multipart MIME type.

This matters because dynamic mode is a supported execution path and may be required in runtimes that do not allow `new Function` or similar code generation.

This may be related to the broader AOT vs non-AOT parity issue discussed in #1753, but this report focuses specifically on `t.File({ type })` content validation.

## Security impact

In `aot: false`, `t.File({ type })` does not appear to guarantee that the uploaded file content matches the allowed type.

An attacker can send arbitrary bytes while declaring a trusted MIME type such as `image/png`.

The impact depends on how the application handles uploaded files after validation. Possible downstream impact includes bypassing “images only” upload rules, storing non-image content under a trusted application domain, passing arbitrary bytes to image processing pipelines, or stored XSS when uploaded content is later served inline in a sensitive origin.

This is not an unconditional RCE or universal XSS. The direct impact depends on downstream storage, serving headers, origin isolation, and content handling. But the validation bypass itself is deterministic in dynamic mode.

## Regression test

The case to lock down is the observable behavior of the public schema API:

```txt
aot: false
t.File({ type: 'image/png' })
SVG bytes uploaded with Content-Type: image/png
expected result: 422
```

Observed today:

```txt
AOT       -> 422
aot:false -> 200
```

<details>
<summary>Reduced regression test idea</summary>

```ts
import { describe, expect, it } from 'bun:test'
import { Elysia, t, type ValidationError } from '../../src'

const spoofedImage = () =>
  new File(
    ['<svg xmlns="http://www.w3.org/2000/svg"><script>alert(1)</script></svg>'],
    'avatar.png',
    { type: 'image/png' }
  )

const app = (aot: boolean) =>
  new Elysia({ aot }).post('/avatar', () => 'ok', {
    body: t.Object({
      file: t.File({ type: 'image/png' })
    })
  })

const uploadSpoofed = (aot: boolean) => {
  const body = new FormData()
  body.append('file', spoofedImage())

  return app(aot).handle(
    new Request('http://localhost/avatar', {
      method: 'POST',
      body
    })
  )
}

describe('t.File type validation parity between AOT and dynamic mode', () => {
  it('rejects a spoofed file in AOT mode', async () => {
    const response = await uploadSpoofed(true)
    expect(response.status).toBe(422)

    const data = (await response.json()) as ValidationError
    expect(data.type).toBe('validation')
  })

  it('rejects a spoofed file in dynamic mode (aot: false)', async () => {
    const response = await uploadSpoofed(false)
    expect(response.status).toBe(422)

    const data = (await response.json()) as ValidationError
    expect(data.type).toBe('validation')
  })
})
```

</details>

The AOT test passes today; the dynamic test fails with `Expected: 422, Received: 200`, and both should pass once parity is restored.

The important point is that the same schema should not provide different validation guarantees depending on the execution mode.

spokodev:
Closes #1922.

`t.File({ type })` / `t.Files({ type })` validates the *actual* file bytes (magic-byte check via `fileType` → `file-type`) in AOT mode, but in dynamic mode (`aot: false`) the same schema only ran the synchronous TypeBox check, which trusts the client-supplied multipart MIME type. A file whose bytes do not match the declared type was accepted in dynamic mode and rejected in AOT mode — the two pipelines disagreed on the same schema.

Repro from the issue (SVG bytes uploaded as `image/png`): AOT → `422`, dynamic → `200`.

### Fix
`compose.ts` (AOT) injects `fileType(c.body.k, extension, 'body.k')` for every File/Files field that declares an extension. `createDynamicHandler` did not. After body validation, iterate the body schema's properties (via the existing `getSchemaProperties`, which also flattens unions) and run the same `fileType(...)` for File/Files fields with an extension — so both modes enforce the identical guarantee using the identical code path and error (`InvalidFileType` → `422`).

### Tests
Added `validate actual file in dynamic mode` to `test/validator/body.test.ts`, mirroring the existing AOT `validate actual file` test but with `new Elysia({ aot: false })`: a real `millenium.jpg` passes (`200`), a spoofed `fake.jpg` is rejected (`422`). Red before the change (dynamic returned `200`), green after. `test/validator/body.test.ts` (57), `test/type-system/files.test.ts` + `formdata.test.ts` (31), and `test/core/dynamic.test.ts` (32) all pass; `tsc --project tsconfig.test.json` is clean.

## Bug Fixes
* File uploads now validate against their declared file type consistently across all application modes.
* File type verification is enforced in dynamic mode, preventing mismatched file uploads.

### DIFF
--- a/src/dynamic-handle.ts
+++ b/src/dynamic-handle.ts
@@ -1,3 +1,4 @@
+import { Kind } from '@sinclair/typebox'
 import { TransformDecodeError } from '@sinclair/typebox/value'
 import type { Context } from './context'
 import { parseCookie } from './cookies'
@@ -10,8 +11,13 @@ import {
 } from './error'
 import type { AnyElysia, CookieOptions } from './index'
 import { parseQuery } from './parse-query'
-import { getSchemaProperties, type ElysiaTypeCheck } from './schema'
+import {
+	getSchemaProperties,
+	unwrapImportSchema,
+	type ElysiaTypeCheck
+} from './schema'
 import type { TypeCheck } from './type-system'
+import { fileType } from './type-system/utils'
 import type { Handler, LifeCycleStore, SchemaValidator } from './types'
 import { hasSetImmediate, redirect, StatusMap, signCookie } from './utils'
 
@@ -615,6 +621,33 @@ export const createDynamicHandler = (app: AnyElysia) => {
 
 						// Zod returns { value: ... } wrapper
 						context.body = decoded?.value ?? decoded
+				}
+
+				// Validate file contents by magic bytes, mirroring the AOT path
+				// (`compose.ts`) which injects `fileType(...)` for File/Files
+				// fields that declare an extension. Without this, dynamic mode
+				// (`aot: false`) only checks the client-supplied MIME type and
+				// accepts a file whose bytes do not match the declared type.
+				if (validator.body) {
+					const bodyProperties = getSchemaProperties(
+						unwrapImportSchema(validator.body.schema)
+					)
+
+					if (bodyProperties)
+						for (const key in bodyProperties) {
+							const property = bodyProperties[key]
+
+							if (
+								property.extension &&
+								(property[Kind] === 'File' ||
+									property[Kind] === 'Files')
+							)
+								await fileType(
+									(context.body as any)?.[key],
+									property.extension,
+									`body.${key}`
+								)
+						}
 				}
 			}
 
--- a/test/validator/body.test.ts
+++ b/test/validator/body.test.ts
@@ -836,6 +836,38 @@ describe('Body Validator', () => {
 		}
 	})
 
+	it('validate actual file in dynamic mode', async () => {
+		const app = new Elysia({ aot: false }).post(
+			'/upload',
+			({ body: { file } }) => file.size,
+			{
+				body: t.Object({
+					file: t.File({
+						type: 'image'
+					})
+				})
+			}
+		)
+
+		{
+			const { request, size } = upload('/upload', {
+				file: 'millenium.jpg'
+			})
+
+			const response = await app.handle(request).then((r) => r.text())
+			expect(+response).toBe(size)
+		}
+
+		{
+			const { request } = upload('/upload', {
+				file: 'fake.jpg'
+			})
+
+			const status = await app.handle(request).then((r) => r.status)
+			expect(status).toBe(422)
+		}
+	})
+
 	it('validate actual file with multiple type', async () => {
 		const app = new Elysia().post(
 			'/upload',


robertokbr:
You're just trying to take credit for someone else's discovery. Do better, man.
https://github.com/elysiajs/elysia/issues/1922

bmtec:
Hey @spokodev, this looks really close to an earlier contribution: https://github.com/elysiajs/elysia/issues/1922
How does yours differ from it? Did you use it as a starting point, by any chance? Just curious =)

spokodev:
@bmtec good question — yes, this PR is the implementation of #1922, which is why it opens with "Closes #1922" and reuses @rafaelcoelhox's repro. The discovery and the diagnosis are theirs; I have not claimed otherwise.

The issue was open with no PR attached, so I took a shot at the actual fix — mirroring the AOT `fileType(...)` call into the dynamic handler so both modes enforce the same guarantee, plus a regression test. Glad to have the maintainers, or @rafaelcoelhox, take it forward however they prefer.

PR #1928, opened 2026-06-23
:bug: fix(type): reject non-decimal numeric strings in IntegerString & NumericEnum

MercuriusDream:
Currently, `t.Integer()` with string inputs (query, path, header) and the string branch of `t.NumericEnum()` validate via `+value` + `isNaN` + `Number.isInteger`. Because unary `+` coerces hex, binary, octal, scientific and whitespace forms to finite numbers, the guard accepts `"0x10"`, `"0b10"`, `"0o17"`, `"1e3"` and `" 5"`, even though none of them are plain decimal integers.

`t.Numeric()` already rejects these through `decimalNumber` in `numeric.ts`. This PR brings the two laxer guards in line with that same regex.

## Breaking

These inputs now return `422` instead of being coerced. `"007"` stays valid (a decimal integer, as `numeric.ts` already does). Valid decimal strings and the number branch of the union are unchanged.

## Tests

Adds `test/type-system/integer-string.test.ts`. Existing validator, type-system and coercion suites are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation for integer-like strings so only plain base-10 formats are accepted.
  * Improved numeric enum input handling to reject non-decimal formats and other invalid numeric values.
  * Existing range checks still apply for valid integer inputs, while malformed numeric strings are now consistently rejected.


MercuriusDream:
@coderabbitai review?

PR #1930, opened 2026-07-03
Improve query parsing and response header handling

assiscabrones:
## Summary

This PR tightens a few hot-path response/query handling details:

- Starts `parseQueryFromURL` and `parseQueryStandardSchema` from the provided query offset instead of scanning the whole URL from index `0`.
- Treats native `Headers` instances as non-empty when they contain entries, so `set.headers = new Headers(...)` is respected consistently.
- Preserves file response defaults (`accept-ranges` and `content-range`) when `set.headers` is a native `Headers` instance.
- Preserves `set` status/headers/cookies for early `FormData` responses in both Web Standard and Bun adapters.
- Adds regression coverage for query offset parsing, native `Headers`, file response headers, and early `FormData` responses.

## Why

`parseQueryFromURL` already accepts a `startIndex`, and the composed handlers pass the query start offset into it. However, the implementation still iterated from the beginning of the URL. For long paths this does extra work on every request with inferred query parsing.

There were also a couple of response mapping edge cases where native `Headers` and early `FormData` responses could lose user-provided response metadata or default file headers.

## Local measurement

A small local micro-benchmark parsing the same long-path URL 300k times improved from roughly:

- Before: `~485ms`
- After: `~142ms`

The largest gain is only expected when the path before `?` is long, but the change also removes unnecessary scanning in the general case.

## Tests

```sh
bun test test/adapter/web-standard/map-early-response.test.ts test/adapter/web-standard/map-response.test.ts test/response/range.test.ts test/response/headers.test.ts test/units/parse-query.test.ts test/validator/query.test.ts
bun run test:types
bun test
```

Full local test run result: `1530 pass, 0 fail`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved custom status, headers, and cookies when returning form data or array-like responses.
  * Improved header handling so existing values aren’t overwritten unexpectedly.
  * Query parsing now respects the provided starting position.
  * Added special handling for `Headers` inputs when checking for empty values.

* **Tests**
  * Expanded coverage for response mapping with custom contexts, `Headers` objects, file-like responses, and query parsing behavior.

### DIFF
--- a/src/adapter/bun/handler.ts
+++ b/src/adapter/bun/handler.ts
@@ -38,7 +38,11 @@ export const mapResponse = (
 				return Response.json(response, set as any)
 
 			case 'ElysiaFile':
-				return handleFile((response as ElysiaFile).value as File, set, request)
+				return handleFile(
+					(response as ElysiaFile).value as File,
+					set,
+					request
+				)
 
 			case 'File':
 				return handleFile(response as File, set, request)
@@ -175,7 +179,11 @@ export const mapEarlyResponse = (
 				return Response.json(response, set as any)
 
 			case 'ElysiaFile':
-				return handleFile((response as ElysiaFile).value as File, set, request)
+				return handleFile(
+					(response as ElysiaFile).value as File,
+					set,
+					request
+				)
 
 			case 'File':
 				return handleFile(response as File, set, request)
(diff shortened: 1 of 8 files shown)


assiscabrones:
@coderabbitai review

PR #1934, opened 2026-07-05
fix(ws): correct inverted response schema validation in createHandleWSResponse

### LINKED ISSUE #1563: WebSocket Zod validator not validating

1.4.16

Darwin 24.6.0 arm64 arm

1.3.3

No

Demo repo at https://github.com/luuktap/elysia-zod-validation-bug

Minimal code snippet
```ts
const app = new Elysia().ws("/zod", {
    query: z.object({
      token: z.uuidv4(),
    }),
    body: z.object({
      type: z.literal("test-one"),
      message: z.string(),
    }),
    open(ws) {
      console.log("Connection opened", ws.data.query.token);
    },
    message(ws, message) {
      console.log("Received message:", message);
      ws.send(`Echo: ${JSON.stringify(message)}`);
    },
    close(ws) {
      console.log("Connection closed", ws.data.query.token);
    },
  })
```

Repro steps
1. Connect any WebSocket client (e.g. Postman) to `ws://localhost:3000/zod?token=56264181-d9c1-4d41-8771-eabdd768662d`
2. Send an invalid message, for example: `{ "hello": true }`

WebSocket server should respond with a validation error.

### What do you see instead?

WebSocket server responds with `Echo: {"hello": true}`.

### Additional information

For easy comparing, I've included a typebox-validated endpoint in my demo repo.

Zod endpoint

Image

Typebox endpoint

Image

Yes

### LINKED ISSUE #1933: WebSocket response schema validation is inverted — valid messages rejected, invalid messages sent unvalidated
### Description
On a `.ws()` route that declares a `response` schema, `createHandleWSResponse` (`src/ws/index.ts`) validates on the wrong condition. `validateResponse` returns truthy on *failure* (TypeBox path: `Check(data) === false`; standard-schema path: returns `.issues`), but the `send()` guard checks `validateResponse(datum) === false` — so the error branch runs when validation passes and is skipped when it fails.

Effect: with a TypeBox `response` schema, a valid handler response is replaced by a `ValidationError` frame sent to the client, and an invalid response is sent as-is; with a standard-schema `response`, invalid responses are sent unvalidated.

### Steps to reproduce
```ts
import { createHandleWSResponse } from './src/ws/index'
import { getSchemaValidator } from './src/schema'
import { t } from './src'

const v = getSchemaValidator(t.Object({ ok: t.Boolean() }), {})
const sent: any[] = []
const ws: any = { send: (x: any) => { sent.push(x); return 0 } }
const handle = createHandleWSResponse(v as any)

handle(ws, { ok: true })  // valid
handle(ws, { bad: 1 })    // invalid
console.log(sent)
```

### Expected
`{ ok: true }` → sent as `{"ok":true}`; `{ bad: 1 }` → replaced by a `ValidationError` frame.

### Actual
`{ ok: true }` → client receives a validation-error frame; `{ bad: 1 }` → sent through as `{"bad":1}`.

### Root cause
`src/ws/index.ts:269` — `if (validateResponse && validateResponse(datum) === false)`. `validateResponse` (defined `src/ws/index.ts:258–266`) returns truthy on failure, so `=== false` double-negates. Regression from commit `23f30f78a` ("fix: #1563 standard schema on websocket"), which wrapped the old `Check(datum) === false` into a predicate but left the `=== false` at the call site. Secondary: the async-generator branch at `src/ws/index.ts:289` validates `first` instead of `first.value`.

### Suggested fix
```ts
// :269
if (validateResponse && validateResponse(datum))
    return ws.send(new ValidationError('message', responseValidator!, datum).message)
// :289 — validate first.value, not first
```

tsushanth:
Fixes #1933.

## Root cause

`createHandleWSResponse` builds a `validateResponse` predicate that returns **truthy on failure**:

- TypeBox path: `(data) => responseValidator.Check(data) === false` — returns `true` when invalid
- Standard-schema path: `(data) => responseValidator.schema['~standard'].validate(data).issues` — returns the issues array (truthy) when invalid

The `send()` guard checked `validateResponse(datum) === false`, which is truthy only when validation **passes** — exactly backwards. Valid messages were replaced by `ValidationError` frames; invalid messages were sent through unvalidated.

Regression from `23f30f78a` ("fix: #1563 standard schema on websocket"), which extracted the old inline `Check(datum) === false` check into a closure but left `=== false` at the call site, double-negating it.

**Second bug in the same block**: the async-generator early-return path validated `first` (the `{value, done}` iterator result object) instead of `first.value` (the handler's actual yielded datum), embedding the wrapper in the `ValidationError` message.

## Fix

1. Remove `=== false` from the `send()` guard — `validateResponse(datum)` alone is truthy on failure.
2. Validate `first.value` instead of `first` in the async-generator branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response validation for websocket interactions so invalid payloads are rejected more reliably.
  * Fixed handling of generator-based responses to validate the returned value correctly and show clearer validation errors when the first result is invalid.

### DIFF
--- a/src/ws/index.ts
+++ b/src/ws/index.ts
@@ -266,7 +266,7 @@ export const createHandleWSResponse = (
 			: undefined
 
 		const send = (datum: unknown) => {
-			if (validateResponse && validateResponse(datum) === false)
+			if (validateResponse && validateResponse(datum))
 				return ws.send(
 					new ValidationError('message', responseValidator!, datum)
 						.message
@@ -286,12 +286,12 @@ export const createHandleWSResponse = (
 			return (async () => {
 				const first = await init
 
-				if (validateResponse && validateResponse(first))
+				if (validateResponse && validateResponse(first.value))
 					return ws.send(
 						new ValidationError(
 							'message',
 							responseValidator!,
-							first
+							first.value
 						).message
 					)
 


tsushanth:
Closing — this has been open without activity long enough that it's more useful to withdraw and let a fresh PR reflect current `main`/`master` than to keep pinging for review on stale state. Happy to reopen or resubmit if there's still interest.

PR #1941, opened 2026-07-12
operation sennen I

SaltyAom:
Codename: [sen'nen](https://youtu.be/3em-J9yYPAo)

sen'nen chan


SaltyAom (commented):
kind of ok, test cases need to remove plan context but ok

PR #1966, opened 2026-08-16
fix: type plain onError return as error status response

zacharywomack8-source:
@algora-pbc /claim #313

---
Type-level fix for #313 (Option to return errors as JSON).

**Problem**
An onError handler that returns a plain value (e.g. `{ error: '...' }` or a string) is currently typed as if the response is `200`. At runtime the response keeps the status of the error that triggered the handler (`400`/`404`/`422`/`500`), so the types lie about what the client receives.

**Change**
- Plain values returned from `onError` are now typed under every built-in error status (`400`, `404`, `422`, `500`) instead of `200`.
- Explicit `status(code, ...)` / `error(code, ...)` returns still resolve to their exact status code via `ExtractErrorFromHandle`.
- Applies to local, scoped, and global `onError` (single handler and array overloads), plus `ErrorHandle` composition in `use`/`guard`/`macro`.

**Implementation**
- Adds `ErrorValueToResponseSchema`, `ElysiaErrorHandlerToResponseSchema`, `ElysiaErrorHandlerToResponseSchemas`, `ElysiaErrorHandlerToResponseSchemaAmbiguous` in `src/types.ts`.
- Swaps the new types into the `onError` overloads and the `ErrorHandle` guards in `src/index.ts`.

**Verification**
- Updated `test/types/lifecycle/soundness.ts` onError expectations: plain returns now produce the error-status union instead of `200`.
- `pnpm test:types` passes.

## Summary by CodeRabbit

* **Improvements**
  * Improved type inference for error-handler responses across local, scoped, and global error handling.
  * Error responses now accurately reflect supported status codes, including 400, 404, 422, and 500.
  * Enhanced type handling for single, multiple, and ambiguous error handlers.
* **Tests**
  * Updated type-safety checks to validate the refined error response behavior.

claim-recheck-marker

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -149,6 +149,9 @@ import type {
 	ElysiaHandlerToResponseSchemas,
 	ExtractErrorFromHandle,
 	ElysiaHandlerToResponseSchemaAmbiguous,
+	ElysiaErrorHandlerToResponseSchema,
+	ElysiaErrorHandlerToResponseSchemas,
+	ElysiaErrorHandlerToResponseSchemaAmbiguous,
 	GuardLocalHook,
 	PickIfExists,
 	SimplifyToSchema,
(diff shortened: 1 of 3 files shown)


SaltyAom:
### I have nothing but my burger, and I want nothing more

In case any person reads this

I'm 100% confident this is an AI-generated PR on how the fix is basically hard-coded status and edit valid test cases to be invalid to fit the narrative

PR #1982, opened 2026-08-26
fix(types): type plain onError returns under error statuses so Eden sees the error body

K4bain:
Fixes #313

## Problem

When an `onError` handler returns a plain value, Elysia keeps the runtime error status (500 for unhandled errors, 422 for validation, etc.) but the response is *typed* as a `200` response. Eden clients then see `{ status: unknown; value: unknown }` instead of the actual body — exactly what @EvHaus describes in the issue.

## What changed

`onError` handlers now infer through a dedicated `ElysiaErrorHandlerToResponseSchema*` type instead of reusing the regular handler mapping:

- A **plain return** is recorded under `400 / 404 / 422 / 500` — the statuses Elysia's built-in error handling assigns at runtime (`PARSE`, `NOT_FOUND`, `VALIDATION`, `INTERNAL_ERROR`). Which one applies depends on the thrown error, so statically it belongs to all of them.
- An explicit **`status(code, value)`** return keeps its exact code and is not duplicated.
- If a plain return coexists with an explicit status that lands on one of those four codes, both bodies are **unioned** under that code (e.g. `404: 'not found' | { failure: boolean }`), rather than intersected into an impossible type.
- Route handlers are untouched — their `200` inference works exactly as before.

This also fixes a latent soundness issue: previously two `onError` sources claiming the same status were intersected (`A & B`) instead of unioned.

## Testing

- New `test/types/lifecycle/on-error-response.ts` covers: plain returns, explicit `status()` returns, mixed unions, and that route handlers keep their `200`
- Updated the three existing `onError` expectations in `soundness.ts` (local/scoped/global) to reflect plain values now landing on error statuses
- `bun run test:types` passes; full `bun test` shows no new failures

@algora-pbc /claim #313

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved response type inference for error handlers across local, scoped, and global configurations.
  - Plain error responses now map to the appropriate error status codes instead of defaulting to success.
  - Explicit status codes, including successful responses, are preserved accurately and are not merged with default error responses.
  - Improved type inference for grouped and guarded error-handling configurations.

- **Tests**
  - Added coverage for plain, explicit, mixed, and ambiguous error-handler responses.

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -2228,6 +2228,85 @@ export type ElysiaHandlerToResponseSchemaAmbiguous<
 				? ElysiaHandlerToResponseSchemas<Schemas>
 				: {}
 
+/**
+ * HTTP statuses an error may resolve to when it reaches an `onError`
+ * handler, as assigned by Elysia's built-in error handling:
+ *
+ * - `400` - `PARSE`
+ * - `404` - `NOT_FOUND`
+ * - `422` - `VALIDATION`
+ * - `500` - unhandled errors (`INTERNAL_ERROR` and unknown errors)
+ *
+ * Which one applies is decided at runtime by the thrown error, so a plain
+ * value returned from an `onError` handler (without an explicit
+ * `status(code, value)`) is recorded under all of them.
+ */
+type DefaultErrorStatusesToResponseSchema<Response> = {
+	400: Response
+	404: Response
+	422: Response
+	500: Response
+}
+
+/**
+ * Like {@link ValueToResponseSchema} but for values returned from an
+ * `onError` handler: explicit `status(code, value)` returns keep their
+ * exact status, while plain returns become the body of the runtime error
+ * status (see {@link DefaultErrorStatusesToResponseSchema}) instead of
+ * being recorded as a `200` response.
+ *
+ * When a plain return coexists with an explicit `status(code, value)`
+ * whose code is one of the default error statuses, both bodies are
+ * unioned under that status.
+ */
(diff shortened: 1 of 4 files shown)


K4bain:
﻿Addressed the CodeRabbit finding: `ErrorValueToResponseSchema` now derives its `Plain` (error-status body) from `Exclude<Value, AnyElysiaCustomStatusResponse>` instead of `Extract200<Value>`, so an explicit `status(200, value)` returned from `onError` is no longer folded into the 400/404/422/500 bodies — it stays under `200` (via `ExtractErrorFromHandle`). Added a type assertion covering `onError(({ status }) => status(200, value))`, which fails against the previous inference and passes now. `tsc --project tsconfig.test.json` is clean.

src/types.ts
```diff
@@ -2228,6 +2228,86 @@ export type ElysiaHandlerToResponseSchemaAmbiguous<
...
+ * - `400` - `PARSE`
+ * - `404` - `NOT_FOUND`
+ * - `422` - `VALIDATION`
+ * - `500` - unhandled errors (`INTERNAL_ERROR` and unknown errors)
+ *
+ * Which one applies is decided at runtime by the thrown error, so a plain
+ * value returned from an `onError` handler (without an explicit
+ * `status(code, value)`) is recorded under all of them.
+ */
+type DefaultErrorStatusesToResponseSchema<Response> = {
```
SaltyAom, line 2244: slop

PR #1993, opened 2026-09-08
fix(types): type plain onError returns under error statuses (#313)

neelinavolu1:
@algora-pbc /claim #313
Fixes #313

## Prerequisites

- [x] I have run `bun test` (1614 pass, 2 pre-existing skips) and `bun run test:types`
- [x] I have added tests to prevent regression in the future that prove my fix is effective or that my feature works
- [x] I understand that I'll write a detailed description of the PR and that it will be reviewed by a human

## Related issue

https://github.com/elysiajs/elysia/issues/313

## Description

Plain objects returned from `onError` keep the thrown error's HTTP status at runtime (`PARSE` 400, `NOT_FOUND` 404, `VALIDATION` 422, unhandled 500), but route metadata typed those bodies as `200`. Eden then treats the custom JSON as a success payload, so clients miss it under `error`.

This is a type-only change. `onError` (and guard `error`) now infer through a dedicated mapper:

- A **plain** return is recorded under 400 / 404 / 422 / 500.
- An explicit **`status(code, value)`** return keeps that code only.
- Overlapping codes union the bodies.
- Route handlers are unchanged — a plain success return is still `200`.

Coverage is in `test/types/lifecycle/on-error.ts`. Existing onError soundness cases were updated so they no longer expect those plain values as `200`.

I have nothing but my burger and I want nothing more

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TypeScript response inference for error handlers.
  - Plain error responses now correctly reflect applicable HTTP statuses, including 400, 404, 422, and 500.
  - Explicitly assigned custom statuses are preserved accurately, including mixed response branches and successful status 200 responses.
  - Error response typing is now consistent across global, scoped, grouped, and guarded error handlers.
- **Tests**
  - Added coverage for error-handler response inference and type safety across lifecycle configurations.

### DIFF
--- a/test/types/lifecycle/on-error.ts
+++ b/test/types/lifecycle/on-error.ts
@@ -1,0 +1,114 @@
+import { Elysia } from '../../../src'
+import { expectTypeOf } from 'expect-type'
+import { Prettify } from '../../../src/types'
+// I have nothing but my burger and I want nothing more
+
+// Issue #313 — a plain onError body is sent with the thrown error's HTTP
+// status (PARSE 400, NOT_FOUND 404, VALIDATION 422, unhandled 500). Eden
+// reads those codes as `error`, so the body must not be typed as 200.
+{
+	const app = new Elysia()
+		.onError(() => ({ failure: 'maintenance' as const }))
+		.get('/', () => {
+			throw new Error('Server is during maintenance')
+		})
+
+	type Response = Prettify<(typeof app)['~Routes']['get']['response']>
+
+	expectTypeOf<Response>().toEqualTypeOf<{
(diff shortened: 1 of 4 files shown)


test/types/lifecycle/on-error.ts
```diff
@@ -0,0 +1,114 @@
+import { Elysia } from '../../../src'
+import { expectTypeOf } from 'expect-type'
+import { Prettify } from '../../../src/types'
+// I have nothing but my burger and I want nothing more
```
SaltyAom, line 4: slop

SaltyAom (changes requested):

PR #2003, opened 2026-09-21
fix(types): expose plain onError returns under error statuses for Eden typing

stffinfcti:
Implements #313.

## Problem
A plain value returned from `onError` keeps the status of the error that triggered it at runtime (`PARSE` 400, `NOT_FOUND` 404, `VALIDATION` 422, `INTERNAL_SERVER_ERROR` 500), but the route response schema recorded it as `200`. Eden therefore could not type `error.value` as the custom JSON error body.

## Change
- Add `ErrorValueToResponseSchema`, `ElysiaErrorHandlerToResponseSchema`, `ElysiaErrorHandlerToResponseSchemas`, `ElysiaErrorHandlerToResponseSchemaAmbiguous` in `src/types.ts`.
- Plain `onError` return values are now recorded under the built-in error statuses instead of `200`.
- Explicit `status(code, ...)` / `error(code, ...)` returns still resolve to their exact status code.
- Applied to every `onError` overload (local/scoped/global, single/array) and to `ErrorHandle` composition in `use`/`guard`/`macro`.

## Tests
- `test/types/index.ts`: Eden-facing route response maps for a plain `onError` return, `status()` preservation, and a global `onError`.
- `test/types/lifecycle/soundness.ts`: updated local/scoped/global `onError` expectations.
- `test/lifecycle/error.test.ts`: plain object from `onError` is returned as JSON with the preserved status.

`bun test` (1615 pass, 2 skip) and `bun run test:types` pass.

/claim #313

## Summary by CodeRabbit

- **New Features**
  - Improved response type inference for error handlers, including plain object responses and explicit status-based responses.
  - Error responses now correctly contribute to route response schemas across global, scoped, and local handlers.

- **Bug Fixes**
  - Plain object responses from error handlers are serialized as JSON while preserving the original error status.

- **Tests**
  - Added coverage for error response serialization and status-aware type inference.


SaltyAom:
see https://github.com/elysiajs/elysia/issues/313#issuecomment-5764097969

edit: also this is the slop
```ts
export type ErrorValueToResponseSchema<Value> = ExtractErrorFromHandle<Value> &
	(Extract200<Value> extends infer R200
		? undefined extends R200
			? {}
			: IsNever<R200> extends true
				? {}
				: {
						400: R200
						404: R200
						422: R200
						500: R200
					}
```

get real

